PR #8, opened 2025-05-13
Auto determine capabilities

kfischer-okarin:
I added logic to auto-determine a server's capabilities based on the protocol features actually used in the server.
- If the server has at least 1 tool or defines a custom `tools_list_handler`, it returns the `tools` capability
- If the server has at least 1 prompt or defines a custom `prompts_list_handler`, it returns the `prompts` capability
- If the server has at least 1 resource or template or defines a custom list handler for either of those it returns the `resources` capability

## Motivation and Context
I found it unnecessary that you have to manually specify the capabilities when they could be actually auto-determined from the server definition.

## How Has This Been Tested?
I tried out a few patterns in my local mcp inspector

## Breaking Changes
This should not be a breaking change since an empty tool list and no capability should make no practical difference to a procotol conform MCP client.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
I was even thinking about completely removing the ability to set or override the capabilities manually but maybe that would have been going too far - and it would have also made testing of `ensure_capability!` more complicated

### DIFF
--- a/lib/model_context_protocol/server.rb
+++ b/lib/model_context_protocol/server.rb
@@ -20,7 +20,8 @@ module ModelContextProtocol
 
     include Instrumentation
 
-    attr_accessor :name, :tools, :prompts, :resources, :server_context, :configuration, :capabilities
+    attr_writer :capabilities
+    attr_accessor :name, :tools, :prompts, :resources, :server_context, :configuration
 
     def initialize(
       name: "model_context_protocol",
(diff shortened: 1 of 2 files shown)


lib/model_context_protocol/server.rb
```diff
@@ -22,7 +22,9 @@ def initialize(message, request, error_type: :internal_error, original_error: ni
 
     include Instrumentation
 
-    attr_accessor :name, :version, :tools, :prompts, :resources, :server_context, :configuration, :capabilities
+
```
topherbullock:
```suggestion
```

auto merge resolving conflicts added this line.

topherbullock (approved):
Good idea! Thanks for this!

I tested with `examples/stdio_server.rb` in MCP inspector.
Removed resources, tools, prompts and got the expected `capabilities` in the `initialize` response.

atesgoral (approved):

atesgoral:
Neat DX idea!

topherbullock:
ty @kfischer-okarin

PR #14, opened 2025-05-21
Fix .gitignore entry for `Gemfile.lock`

andyw8:
The `Gemfile.lock` entry has a `.` in front which prevents it from being ignored.


topherbullock (approved):

topherbullock:
🤦🏻 ty @andyw8!

PR #24, opened 2025-05-26
Update gemspec to reflect `mcp` naming

topherbullock:
## Motivation and Context

the gem will be published to RubyGems under https://rubygems.org/gems/mcp as `mcp`, so gemspec, lib structure and naming should reflect this.

## How Has This Been Tested?

moved files, renamed constants, ran tests; CI is green

## Breaking Changes
none. `ModelContextProtocol` is aliased to `MCP` to allow backward compatibility if the gem has been leveraged using git instead of RubyGems by intrepid users.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,29 @@
 
 A Ruby gem for implementing Model Context Protocol servers
 
+## Installation
+
+Add this line to your application's Gemfile:
+
+```ruby
+gem 'mcp'
+```
+
+And then execute:
+
+```bash
+$ bundle install
+```
+
+Or install it yourself as:
+
+```bash
+$ gem install mcp
+```
+
 ## MCP Server
 
-The `ModelContextProtocol::Server` class is the core component that handles JSON-RPC requests and responses.
+The `MCP::Server` class is the core component that handles JSON-RPC requests and responses.
 It implements the Model Context Protocol specification, handling model context requests and responses.
 
 ### Key Features
@@ -44,19 +64,17 @@ requests.
 You can use the `Server#handle_json` method to handle requests.
 
 ```ruby
-module ModelContextProtocol
-  class ApplicationController < ActionController::Base
+class ApplicationController < ActionController::Base
 
-    def index
-      server = ModelContextProtocol::Server.new(
-        name: "my_server",
-        version: "1.0.0",
-        tools: [SomeTool, AnotherTool],
-        prompts: [MyPrompt],
-        server_context: { user_id: current_user.id },
-      )
-      render(json: server.handle_json(request.body.read).to_h)
-    end
+  def index
+    server = MCP::Server.new(
+      name: "my_server",
+      version: "1.0.0",
+      tools: [SomeTool, AnotherTool],
+      prompts: [MyPrompt],
+      server_context: { user_id: current_user.id },
+    )
+    render(json: server.handle_json(request.body.read))
   end
 end
 ```
@@ -67,11 +85,11 @@ If you want to build a local command-line application, you can use the stdio tra
 
 ```ruby
 #!/usr/bin/env ruby
-require "model_context_protocol"
-require "model_context_protocol/transports/stdio"
+require "mcp"
+require "mcp/transports/stdio"
 
 # Create a simple tool
-class ExampleTool < ModelContextProtocol::Tool
+class ExampleTool < MCP::Tool
   description "A simple example tool that echoes back its arguments"
   input_schema(
     properties: {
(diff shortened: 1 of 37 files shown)


koic:
I think that `model_context_protocol` (`ModelContextProtocol` module) might be a clearer name, rather than `mcp` (`MCP` module). The name appears to be effectively available:
https://rubygems.org/gems/model_context_protocol

In case publishing is not possible for some reason, contacting RubyGems would be advisable:
support@rubygems.org

From a long-term perspective, a more descriptive name may be preferable to an abbreviation. A unique name, rather than an abbreviated one that might cause conflicts, is likely to be clearer for users.

topherbullock:
> In case publishing is not possible for some reason, contacting RubyGems would be advisable: [support@rubygems.org](mailto:support@rubygems.org)

we've worked with RubyGems to claim https://rubygems.org/gems/mcp - MCP is established as an acronym in the sdks for other languages, so `mcp` as a gem name is preferred.

koic:
I see. I might have been too influenced by an LSP SDK gem:
https://rubygems.org/gems/language_server-protocol

Indeed, aligning with SDKs in other languages could be an option. Thank you for the explanation!

atesgoral (approved):

atesgoral:
~~Let's Fucking Go!~~ LFG!

atesgoral (approved):

test/model_context_protocol/server_test.rb
```diff
@@ -61,7 +61,7 @@ class ServerTest < ActiveSupport::TestCase
       )
     end
 
-    # https://spec.modelcontextprotocol.io/specification/2025-03-26/basic/utilities/ping/#behavior-requirements
+    # https://spec.MCP.io/specification/2025-03-26/basic/utilities/ping/#behavior-requirements
```
kfischer-okarin, line 64: I think this was one replace too much

lib/mcp.rb
```diff
@@ -0,0 +1,42 @@
...
+    attr_reader :audience, :priority
+
+    def initialize(audience: nil, priority: nil)
+      @audience = audience
+      @priority = priority
+    end
+  end
+end
+
+ModelContextProtocol = MCP
```
kfischer-okarin: Is this like an alias for easy migration?

One could maybe even do some meta programming powered warnings I don't know like

```rb
module ModelContextProtocol
  def self.const_missing(name)
    warn('ModelContextProtocol is deprecated. Use MCP instead')
    MCP.const_get(name)
  end
end
```

test/model_context_protocol/server_test.rb
atesgoral: Reverted!

lib/mcp.rb
kfischer-okarin: But then again on second thought - its still a pre-1.0 release (and not yet even any official push to rubygems) so breaking APIs without warnings should be fine....

I was just wondering if this alias still has any good reason to exist now that the official name has become the desired short module name `MCP`

README.md
```diff
@@ -67,8 +87,8 @@ If you want to build a local command-line application, you can use the stdio tra
 
 ```ruby
 #!/usr/bin/env ruby
-require "model_context_protocol"
-require "model_context_protocol/transports/stdio"
+require "mcp"
+require "mcp/transports/stdio"
 
 # Create a simple tool
 class ExampleTool < ModelContextProtocol::Tool
```
koic: It looks like there are still some places in the example code that need to be updated to use `MCP`. e.g.,

```suggestion
class ExampleTool < MCP::Tool
```

lib/mcp.rb
koic: Yeah, since the gem hasn't been published to RubyGems yet, there may be some judgment involved in how much to care about this. That said, if a warning is used, specifying `uplevel: 1` would be more user-friendly.

```diff
-warn('ModelContextProtocol is deprecated. Use MCP instead')
+warn('ModelContextProtocol is deprecated. Use MCP instead.', uplevel: 1)
```

topherbullock: Good points @kfischer-okarin @koic,  this doesn't have much use other than helping any intrepid users who have used the gem from git source before publishing to RubyGems.

Shopify has an internally published version of the gem as `ModelContextProtocol` but we can manage a s/ModelContextProtocol/MCP/

I'll drop the alias; hard to remove once it's out there on RubyGems without a breaking release.

atesgoral (approved):

PR #27, opened 2025-05-28
Refactor in preparation for client support

jcat4:
## Motivation and Context

We want to implement client support for this lib. Currently, all the code isn't clearly organized between shared and server-specific code. To be more inline with other sdk's, I've organized the code into /server and /shared dirs.

Now we can relatively easily add a new /client dir and implement client support there.

## How Has This Been Tested?

We tested on one of our internal applications, which has an MCP server. We used it from cursor using this version, and everything works just fine.

## Breaking Changes

`MCP::Transports::StdioTransport` is now `MCP::Server::Transports::StdioTransport`

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

- Did I miss anything?
- Is the `1.0.0-preview` appropriate? I was thinking the official `1.0.0` could be reserved for the client support. Or this could be `1.0.0` and the client support can be `1.1.0`? Since there _are technically_ breaking changes, I don't want to do a minor bump.

PR #28, opened 2025-05-28
Add basic http client support

jcat4:
## Motivation and Context

Follow-up to https://github.com/modelcontextprotocol/ruby-sdk/pull/27

This adds the ability to build MCP clients with the ruby sdk. I've started with a basic HTTP client. We can add other things (i.e. streamable HTTP) later.

## Types of changes

- [ ] Documentation update

## Checklist


jcat4:
Decided to build in auth, going to draft to test it

jcat4:
Auth is tested and working

topherbullock:
@jcat4 can you rebase on main and resolve conflicts? seems to be largely the `model_context_protocol => mcp` rename

vickyonit:
Thank you for your work on this, @jcat4. We're looking forward to seeing it merged. Is there an expected timeline for this?

jcat4:
> @jcat4 can you rebase on main and resolve conflicts? seems to be largely the `model_context_protocol => mcp` rename

Sorry, lost track of this.

I thought we wanted to go with [the pluggable approach](https://github.com/modelcontextprotocol/ruby-sdk/tree/add-client-with-pluggable-transports)? If so, I can close this

Or I can try to rework this to be closer to what y'all envisioned!

atesgoral:
@jcat4 @topherbullock Now that #27 is in, could we revisit this and get a client in there? 🙏

lib/model_context_protocol/client/http.rb
```diff
@@ -0,0 +1,107 @@
...
+        ::ModelContextProtocol::Client::Tools.new(response)
+      end
+
+      def call_tool(tool:, input:)
+        response = make_request(
+          method: "tools/call",
+          params: { name: tool.name, arguments: input },
+        ).body
+
+        response.dig("result", "content", 0, "text")
```
juharris: Sorry if I'm missing something, but what about other types of tool responses? Should we assume that the text property needs to be used?

model_context_protocol.gemspec
```diff
@@ -27,6 +27,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
+  spec.add_dependency("faraday", ">= 2.0")
```
juharris: Let's make this an optional dependency?

Are we sure that we want to combine both the server and client in one gem? Sure it's super handy for us to develop and distribute, but I assume most people will either be making a server or a client and rarely both in the same project. So I would rather keep the dependencies minimal depending on what the developer is doing. Can we separate the two somehow? I looked at a few of the other SDKs and it seemed like the Python, Rust, and .NET versions are doing some of the things to not force you to have all of the dependencies when you install their SDK. So I think there are a few ways to do this in Ruby. An optional dependency is one way, but I'm open to other suggestions.

lib/model_context_protocol/client/http.rb
atesgoral: That's true. `text` is not a mandatory content chunk to have.

jcat4:
> @jcat4 @topherbullock Now that #27 is in, could we revisit this and get a client in there? 🙏

I'll make some time this week to knock this out

model_context_protocol.gemspec
jcat4: That's a very good point. I can patch this to be optional and clarify that in the docs

jcat4: Looks like there's some discussion on this, but no progress yet
https://github.com/rubygems/rfcs/issues/27

What do you think about just having a comment/readme saying _your project_ needs to have faraday installed to use the client HTTP transport. And the client code can lazy-load faraday and raise a verbose error about it not being included if it can't find it?

jcat4: Took a crack at it [here](https://github.com/modelcontextprotocol/ruby-sdk/pull/28/commits/72e07aa78887b68ea81a2262faec8d9b31f3ce15)

juharris: Awesome! Yeah I think that's fine. It's sort of what the Python version does: `pip install "mcp[cli]"` https://github.com/modelcontextprotocol/python-sdk?tab=readme-ov-file#installation
https://github.com/modelcontextprotocol/python-sdk/blob/main/pyproject.toml#L38-L41

Rust also has "features":
https://github.com/modelcontextprotocol/rust-sdk/blob/main/crates/rmcp/Cargo.toml#L77-L81

.NET is split into several SDKs: https://github.com/modelcontextprotocol/csharp-sdk/tree/main?tab=readme-ov-file#packages

README.md
```diff
@@ -10,6 +10,21 @@ Add this line to your application's Gemfile:
 gem 'mcp'
 ```
 
+### Optional Dependencies
+
+The MCP gem has different dependency requirements depending on your use case:
+
+**For Server-only usage:**
```
juharris: I think we should explain more clearly what server and client mean here. For example, in .NET, there are more details: https://github.com/modelcontextprotocol/csharp-sdk/tree/main?tab=readme-ov-file#packages

I would make it more like a simple decision tree:

If you want to host an MCP tool in a service, do this: ...

If you want to call MCP tools and be a client, do this: ...

lib/mcp/client/http.rb
```diff
@@ -0,0 +1,115 @@
...
+      end
+
+      def require_faraday!
+        require "faraday"
+      rescue LoadError
+        raise LoadError, "The 'faraday' gem is required to use the MCP client HTTP transport. " \
+          "Add it to your Gemfile: gem 'faraday', '>= 2.0'"
+      end
+
+      def make_request(method:, params: nil)
```
juharris: It does more than only "make" the request.

```suggestion
      def send_request(method:, params: nil)
```

lib/mcp/client/http.rb
```diff
@@ -0,0 +1,115 @@
...
+        ::MCP::Client::Tools.new(response)
+      end
+
+      def call_tool(tool:, input:)
+        response = make_request(
+          method: "tools/call",
+          params: { name: tool.name, arguments: input },
+        ).body
+
+        response.dig("result", "content", 0, "text")
```
juharris: I think we should not assume text, as mentioned here: https://github.com/modelcontextprotocol/ruby-sdk/pull/28#discussion_r2239851160

I would consider generalizing this somehow to make it easy to get any property. Maybe expose a method that doesn't `dig` at all. Then a few other convenience methods can do something like `call_tool(...).dig(...)`

juharris: For instance, on my team, we're considering adding new custom properties to tool responses and we want to be able to access them.

juharris: We're going to use `"result"`'s `"structuredContent"`. https://modelcontextprotocol.io/specification/2025-06-18/server/tools#output-schema

jcat4: Very good input! I can make text a default arg and give callers the ability to provide their own property. Something like

```ruby
call_tool(tool:, input:, property: "text")
  ...
  
  response.dig("result", "content", 0, property)
end
```


That should suffice, ya? Or do we need more complex options?

jcat4: Oooor we could just return the all the results and let callers be responsible for:
- picking which response(s) they want
- pulling the properties they want from each

That feels a little more flexible

README.md
jcat4: Took a crack at separating what we have into Server/Client

I'm considering if it's worth breaking the README up – it's getting large. But I don't like it when libraries make me dig for the thing I need by fragmenting everything. Open to thoughts here

README.md
```diff
@@ -216,7 +218,7 @@ $ ruby examples/stdio_server.rb
 {"jsonrpc":"2.0","id":"2","method":"tools/list"}
 ```
 
-## Configuration
+### Configuration
```
jcat4, line 223: At the moment, this is server-specific. If we have this patch a client config too (or stop having the config scoped to just the server), we can move this somewhere else in the README

lib/model_context_protocol/client/http.rb
jcat4: Changed this to just return all response content and leave it up to callers to pluck what they need as suggested [here](https://github.com/modelcontextprotocol/ruby-sdk/pull/28#discussion_r2243840741)

I didn't write any wrapper classes since the shape of the content can be so open-ended

jcat4:
Unsure if we still feel strongly about going with a pluggable approach, and haven't had any feedback on that yet. Will put this as-is in for review

But please let me know if we should go somewhere else with the interface! I want us to get as close to right as possible on our first release

jcat4:
Unsure if we still feel strongly about going with a pluggable approach, and haven't had any feedback on that yet. Will put this as-is in for review

But please let me know if we should go somewhere else with the interface! I want us to get as close to right as possible on our first release

jcat4:
I've tested the new code in one of our local apps, and it's working as expected.

jcat4:
Spoke with @juharris, and we decided a pluggable approach would probably be better. The primary reason is it will allow folks to implement and pass their own custom transport layers (and open the door for other gems to add more, too!). The first step might be a bit manual (create your transport, create your client by passing your transport, send messages to your client), but we can iterate later with convenience wrappers that won't break existing code

Overall, it seems like the safer bet. What we have is functional, but maybe not flexible enough for a public gem. Will begin work on this again next week when I have time, I think I can knock something out relatively quickly

.gitignore
```diff
@@ -8,3 +8,6 @@
 /spec/reports/
 /tmp/
 Gemfile.lock
+
+# Mac stuff
+.DS_Store
```
atesgoral: ❤️

atesgoral:
@jcat4 Thanks for doing this!

lib/mcp/client/tools.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+
+      def all
+        tools
+      end
+
+      private
+
+      def tools
+        @tools ||= @response.dig("result", "tools")&.map { |tool| Tool.new(tool) } || []
```
jcat4: Do you think this is flexible enough? Working on the pluggable transport layers, and a part of the client interface is I'd like all transport layers to return instances of this, or at least something with the same interface. My goal is for `MCP::Client.new(transport: any_transport).tools` with any transport layer to return the same shaped thing. Having wildly different types of responses based on the transport layer feels very wrong to me

Maybe this object is unnecessary, and the interface should instead be returning a dumb array of `MCP::Client::Tool`s? And `MCP::Client::Tool` could accept kwargs instead of digging in a hash? Just to give maximum flexibility for different transport layers

lib/mcp/client/http.rb
```diff
@@ -0,0 +1,115 @@
+# frozen_string_literal: true
+
+module MCP
+  module Client
+    class Http
+      DEFAULT_VERSION = "0.1.0"
```
jcat4: I don't know what purpose this serves. I think I copied some other code in the app. If there are no strong feelings, I'm going to drop this and the version attr

lib/mcp/client.rb
```diff
@@ -0,0 +1,68 @@
...
+    # @return [Object] The result of the tool call, as returned by the transport.
+    #
+    # @example
+    #   tool = client.tools.first
+    #   result = client.call_tool(tool: tool, input: { foo: "bar" })
+    #
+    # @note
+    #   The exact requirements for `input` are determined by the transport layer in use.
+    #   Consult the documentation for your transport (e.g., MCP::Client::HTTP) for details.
+    def call_tool(tool:, input: nil)
```
jcat4: could `**kwargs` this is we want to let folks pass their own special stuff for custom transport layers, too

jcat4: Was torn on this while building. This class is really just a simple wrapper that delegates everything to the transport layer. Maybe that's just coincidental, and _eventually_ these methods will do more that will make the abstraction worth it

lib/mcp/client/tools.rb
jcat4: got rid of this class entirely

lib/mcp/client/http.rb
jcat4: dropped, pls let me know if this should be added back and why

lib/mcp/client.rb
```diff
@@ -0,0 +1,68 @@
...
+    #   This allows the client to list available tools and invoke tool calls via the transport.
+    #
+    # @example
+    #   transport = MCP::Client::HTTP.new(url: "http://localhost:3000")
+    #   client = MCP::Client.new(transport: transport)
+    #
+    # @note
+    #   The transport does not need to be a specific class, but must implement:
+    #     - #tools
+    #     - #call_tool(tool:, input:)
```
jcat4: I didn't see other YARD docs in this gem, but felt it might be useful to document ducktypes and whatnot. I want this to be easy to understand

lib/mcp/client.rb
```diff
@@ -0,0 +1,68 @@
...
+    #
+    # @return [Array<MCP::Client::Tool>] An array of available tools.
+    #
+    # @example
+    #   tools = client.tools
+    #   tools.each do |tool|
+    #     puts tool.name
+    #   end
+    def tools
+      @tools ||= transport.tools
```
juharris: I wouldn't memoize tools. The server might change the list of available tools and it should be up to apps how much they wanna cache that.

README.md
```diff
@@ -583,6 +585,73 @@ end
...
+### HTTP Transport Layer
+
+You'll need to add `faraday` as a dependency to use the HTTP transport layer.
+
+```ruby
+gem 'mcp'
+gem 'faraday', '>= 2.0'
+```
+
+The `MCP::Client::Http` class provides a simple HTTP client for interacting with MCP servers:
```
juharris: We gotta update these examples to use the transport now, right?
```suggestion
The `MCP::Client` class provides a simple HTTP client for interacting with MCP servers:
```

lib/mcp/client/http.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
+
+module MCP
+  class Client
+    class HTTP
```
juharris, line 5: So HTTP is one of the transports? Let's document that here and update the examples in the README.

lib/mcp/client/http.rb
```diff
@@ -0,0 +1,118 @@
...
+      attr_reader :url
+
+      def initialize(url:, headers: {})
+        @url = url
+        @headers = headers
+      end
+
+      def tools
```
juharris: I think the transports should be more low level. They should just make requests and return JSON RPC responses. They shouldn't even really know about "tools", maybe not yet.
So the Client would create requests and parse responses. Imagine that we have to re-implement this using SSEs or WebSockets. What code would be re-use in the client?

lib/mcp/client.rb
jcat4: I think that's fair! Was torn on this.

I'll document that this isn't cached to hopefully prevent folks from accidentally making multiple calls.

README.md
jcat4: Oh right, yeah I'll fix this up

jcat4: Donezo! Also more clearly documented the HTTP transport layer and building your own transport layer.

Lmk what you think

lib/mcp/client/http.rb
jcat4: Correct if I'm misunderstanding. I thought the transport layer should tell the client how to perform _all_ communications with the server, including asking for tools. So if you need to communicate with a server with transport layer Y, that means both calling tools and _asking for tools_ should be done using transport layer Y. So then it'd make sense for the transport layer to provide those instructions, right?

Or are you suggesting that we should instead have the transport layers just implement something like `send_request(method: params:, headers:)` and then that could act as a template method for things like `call_tool` and `tools`?

jcat4: If we assume that no matter what transport layer we use, the response should come back as JSON with a consistent format, we could def go that direction and then handle rolling up the `Tool` objects and calling tools + returning data for them. I assume the MCP spec would guarantee this for us?

topherbullock: the transport will always be communicating back to the server using JSON-RPC formatted requests that map to MCP operations. these always follow a predictable pattern, so I don't see why the client transport would need to know specifically about how to format the JSON RPC request for a `tools/list`. 

the client transport just needs to know how to send generic blobs of JSON RPC to the server and read the respons; a higher level abstraction knows how to build `tools/list`, etc.

lib/mcp/client/http.rb
```diff
@@ -0,0 +1,118 @@
...
+
+      def send_request(method:, params: nil)
+        client.post(
+          "",
+          {
+            jsonrpc: "2.0",
+            id: request_id,
+            method:,
+            params:,
+            mcp: { jsonrpc: "2.0", id: request_id, method:, params: }.compact,
```
juharris: What's with the duplication in the `mcp` prop?

lib/mcp/client/http.rb
jcat4: I gotcha! I'll cook up something

lib/mcp/client.rb
jcat4: No longer relevant now that transports just implement a template method. The abstraction makes sense now 👍

lib/mcp/client/http.rb
jcat4: good question. I don't see this in the spec, not sure why I added it tbh. I've removed it and confirmed it still works on one of our internal apps

lib/mcp/client/http.rb
jcat4: I've pushed a change that I think does what y'all suggested. Lmk what you think!

lib/mcp/client/http.rb
```diff
@@ -0,0 +1,87 @@
...
+            faraday.headers[key] = value
+          end
+        end
+      end
+
+      def require_faraday!
+        require "faraday"
+      rescue LoadError
+        raise LoadError, "The 'faraday' gem is required to use the MCP client HTTP transport. " \
+          "Add it to your Gemfile: gem 'faraday', '>= 2.0'"
```
topherbullock:
```suggestion
        raise LoadError, "The 'faraday' gem is required to use the MCP client HTTP transport. " \
          "Add it to your Gemfile: gem 'faraday', '>= 2.0'" \ "See https://rubygems.org/gems/faraday for more details."
```

might be good to point folks to the rubygems page for faraday

lib/mcp/client.rb
```diff
@@ -0,0 +1,96 @@
...
+    # @return [Object] The result of the tool call, as returned by the transport.
+    #
+    # @example
+    #   tool = client.tools.first
+    #   result = client.call_tool(tool: tool, input: { foo: "bar" })
+    #
+    # @note
+    #   The exact requirements for `input` are determined by the transport layer in use.
+    #   Consult the documentation for your transport (e.g., MCP::Client::HTTP) for details.
+    def call_tool(tool:, input: nil)
```
topherbullock:
```suggestion
    def call_tool(tool:, arguments: nil)
```

lib/mcp/client.rb
```diff
@@ -0,0 +1,96 @@
...
+      response = transport.send_request(request: {
+        jsonrpc: JSON_RPC_VERSION,
+        id: request_id,
+        method: "tools/call",
+        params: { name: tool.name, arguments: input },
```
topherbullock:
```suggestion
        params: { name: tool.name, arguments: arguments },
```

topherbullock (approved):
LGTM - one nit about `call_tool`'s arguments. `input` --> `arguments` makes it more obvious what is being passed in to the tools/call request.

topherbullock:
@jcat4 can you fix up the suggestions and squash your commits pls? Thanks for pushing this through!

lib/mcp/client.rb
jcat4: fair enough, will patch that up!

mcp.gemspec
```diff
@@ -29,4 +29,6 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency("json_rpc_handler", "~> 0.1")
   spec.add_dependency("json-schema", ">= 4.1")
+
+  # Faraday is required for the client HTTP transport layer
```
koic: Since it seems that the policy is for users to add `faraday` to their Gemfile, this comment appears unnecessary.

.gitignore
koic: Since this is a user-specific setting, I think it should be defined in the developer's `~/.gitignore` rather than in the repository.

Gemfile
```diff
@@ -22,3 +22,8 @@ gem "activesupport"
 gem "debug"
 gem "rake", "~> 13.0"
 gem "sorbet-static-and-runtime"
+
+group :test do
+  gem "webmock"
+  gem "faraday", ">= 2.0"
+end
```
koic: The code in L9–L11 also looks like it is for testing. It seems better to move them and sort in ASCII order.

```suggestion
group :test do
  gem "faraday", ">= 2.0"
  gem "minitest", "~> 5.1", require: false
  gem "minitest-reporters"
  gem "mocha"
  gem "webmock"
end
```

README.md
```diff
@@ -595,3 +689,4 @@ Releases are triggered by PRs to the `main` branch updating the version number i
 1. **Merge your PR to the main branch** - This will automatically trigger the release workflow via GitHub Actions
 
 When changes are merged to the `main` branch, the GitHub Actions workflow (`.github/workflows/release.yml`) is triggered and the gem is published to RubyGems.
+
```
koic:
```suggestion
```

README.md
```diff
@@ -583,6 +585,98 @@ end
...
+  end
+end
+```
+
+### HTTP Transport Layer
+
+Use the `MCP::Client::Http` transport to interact with MCP servers using simple HTTP requests.
+
+The HTTP client supports:
+- Tool listing via the `tools/list` method
```
koic: That's just my two cents. Can you insert a blank line?

```suggestion
The HTTP client supports:

- Tool listing via the `tools/list` method
```

lib/mcp/client.rb
```diff
@@ -0,0 +1,96 @@
...
+    # @return [Array<MCP::Client::Tool>] An array of available tools.
+    #
+    # @example
+    #   tools = client.tools
+    #   tools.each do |tool|
+    #     puts tool.name
+    #   end
+    def tools
+      response = transport.send_request(request: {
+        jsonrpc: JSON_RPC_VERSION,
```
koic: How about using the constant defined in the dependency gem instead of defining one independently?

```suggestion
        jsonrpc: JsonRpcHandler::Version::V2_0,
```

README.md
```diff
@@ -583,6 +585,98 @@ end
...
+The `MCP::Client` class provides an interface for interacting with MCP servers.
+Clients are initialized with a transport layer instance that handles the low-level communication mechanics.
+
+## Transport Layer Interface
+
+If the transport layer you need is not included in the gem, you can build and pass your own instances so long as they conform to the following interface:
+
+```ruby
+class CustomTransport
+  # Sends a JSON-RPC request to the server and returns the raw response
```
koic:
```suggestion
  # Sends a JSON-RPC request to the server and returns the raw response.
```

README.md
```diff
@@ -583,6 +585,98 @@ end
...
+```ruby
+http_transport = MCP::Client::HTTP.new(url: "https://api.example.com/mcp")
+client = MCP::Client.new(transport: http_transport)
+
+# List available tools
+tools = client.tools
+tools.each do |tool|
+  puts "Tool: #{tool.name}"
+  puts "Description: #{tool.description}"
+  puts "Input Schema: #{tool.input_schema}"
```
koic:
```suggestion
  puts <<~TOOL_INFORMATION
    Tool: #{tool.name}
    Description: #{tool.description}
    Input Schema: #{tool.input_schema}
  TOOL_INFORMATION
```

koic:
Thank you for the client implementation. I've left some minor feedback. Since it doesn’t seem to be squashed yet, please squash your commits into one at the end.

jcat4:
> Since this is a user-specific setting, I think it should be defined in the developer's ~/.gitignore rather than in the repository.

@koic My view: if we never want a file committed, the repo should enforce it so we don’t rely on each contributor’s local setup. Is there a strong reason to prefer user-level ignores here?

koic:
Because these are specific to the user's development environment (such as OS or editor) and not specific to the Ruby SDK. For more details, please refer to the discussion in the following closed PRs:
https://github.com/rails/rails/pulls?q=is%3Apr+.DS_Store+is%3Aclosed+

jcat4:
> Because these are specific to the user's development environment (such as OS or editor) and not specific to the Ruby SDK. For more details, please refer to the discussion in the following closed PRs: https://github.com/rails/rails/pulls?q=is%3Apr+.DS_Store+is%3Aclosed+

Thanks for this! Hmmm... I do see that's how rails operates. If this is commonly how other open source ruby repos operate, I can make this change to be more inline.

But I do wanna say, I still personally disagree, as I think this change could avoid friction. Those 14 PRs in rails where this convo had to happen are examples of friction that could be avoided.

If you have any docs or anything on hand, would love to know more about _why_ we want to have a platform-agnostic .gitignore!

koic:
I'm personally convinced by the reasoning in the Rails repository, so I don't have any additional documentation beyond that. In any case, this doesn't seem to be the main focus of this PR. This point can serve as a starting place for the PR itself. I do have a bit of feedback, so I'll leave some additional comments.

lib/mcp/client/tool.rb
```diff
@@ -0,0 +1,16 @@
+# typed: false
```
koic: This pragma appears to be unnecessary.

```suggestion
```

test/mcp/client_test.rb
```diff
@@ -0,0 +1,55 @@
...
+            { "name" => "tool1", "description" => "tool1", "inputSchema" => {} },
+            { "name" => "tool2", "description" => "tool2", "inputSchema" => {} },
+          ],
+        },
+      }
+
+      # Only checking for the essential parts of the request
+      transport.expects(:send_request).with do |args|
+        args.dig(:request, :method) == "tools/list" &&
+          args.dig(:request, :jsonrpc) == "2.0"
```
koic: Using pattern matching instead of `dig` method can eliminate key redundancy and make the expected structure clearer.

```suggestion
        args in { request: { method: "tools/list", jsonrpc: "2.0" } }
```

test/mcp/client_test.rb
```diff
@@ -0,0 +1,55 @@
...
+      mock_response = {
+        "result" => { "content" => "result" },
+      }
+
+      # Only checking for the essential parts of the request
+      transport.expects(:send_request).with do |args|
+        args.dig(:request, :method) == "tools/call" &&
+          args.dig(:request, :jsonrpc) == "2.0" &&
+          args.dig(:request, :params, :name) == "tool1" &&
+          args.dig(:request, :params, :arguments) == arguments
```
koic: Ditto.

```suggestion
        args in { request: { method: "tools/call", jsonrpc: "2.0", params: { name: "tool1", arguments: arguments } } }
```

lib/mcp/client/tool.rb
jcat4: I can't recall where this comes from (I think a rubocop rule that's probably since been removed in this repo)

We do have these in other files, so may be worth going in and cleaning those up in another PR too!

test/mcp/client_test.rb
jcat4: Oh neat! Was not familiar with this syntax

koic (approved):

jcat4:
I had merge conflicts, had to rebase, resolve, and squash. Nothing of substance changed since last approval

**EDIT:** nvm, [this part](https://github.com/modelcontextprotocol/ruby-sdk/pull/28#discussion_r2322371525) of the README changed, and found some stale YARD docs in the client

README.md
```diff
@@ -612,6 +614,101 @@ server = MCP::Server.new(
...
+
+Use the `MCP::Client::Http` transport to interact with MCP servers using simple HTTP requests.
+
+The HTTP client supports:
+
+- Tool listing via the `tools/list` method
+- Tool invocation via the `tools/call` method
+- Automatic JSON-RPC 2.0 message formatting
+- UUID request ID generation
+- Setting headers for things like authorization
```
jcat4: I'm actually gonna move this, this is now the responsibility of the client

README.md
```diff
@@ -612,6 +614,102 @@ server = MCP::Server.new(
...
+    # - WebSocket: Send message over WebSocket
+    # - stdio: Write to stdout, read from stdin
+    # - etc.
+  end
+end
+```
+
+### HTTP Transport Layer
+
+Use the `MCP::Client::Http` transport to interact with MCP servers using simple HTTP requests.
```
koic:
```suggestion
Use the `MCP::Client::HTTP` transport to interact with MCP servers using simple HTTP requests.
```

jcat4: ah yes

koic (approved):

koic:
This MCP client implementation can serve as a starting point and be improved as needed. In any case, no further conflicts should arise with this PR. Thank you for working on it!

PR #33, opened 2025-05-30
Implement Streamable HTTP transport

### LINKED ISSUE #4: support the streamable part of Streamable HTTP

Ginja:
Implements a rack-compatible Streamable HTTP transport with SSE support. As well as add notification support for the server with the following methods:

- `notify_tools_list_changed()` - Send a notification when the tools list changes
- `notify_prompts_list_changed()` - Send a notification when the prompts list changes  
- `notify_resources_list_changed()` - Send a notification when the resources list changes

Closes #4

## Motivation and Context
To fulfill the Streamable HTTP transport specification as described [here](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/2025-03-26/basic/transports.mdx#streamable-http).

## How Has This Been Tested?
Local tests only (see `examples/`). Working on implementing it with my MCP server that's a WIP.

## Breaking Changes
Should introduce no breaking changes as using the StreamableHTTP transport is optional

## Types of changes

- [x] Documentation update

## Checklist

## Local Tests

```
Started with run options --seed 59936

  132/132: [============================================================================================================================================================================================] 100% Time: 00:00:01, Time: 00:00:01

Finished in 1.21315s
132 tests, 405 assertions, 0 failures, 0 errors, 0 skips
```


README.md
```diff
@@ -33,6 +33,7 @@ It implements the Model Context Protocol specification, handling model context r
 - Manages tool registration and invocation
 - Supports prompt registration and execution
 - Supports resource registration and retrieval
+- Supports stdio & StreamableHTTP (including SSE) transports
```
atesgoral:
```suggestion
- Supports stdio & Streamable HTTP (including SSE) transports
```

examples/README.md
```diff
@@ -0,0 +1,174 @@
...
+A simple server that communicates over standard input/output. This is useful for desktop applications and command-line tools.
+
+**Usage:**
+```bash
+ruby examples/stdio_server.rb
+{"jsonrpc":"2.0","id":0,"method":"tools/list"}
+```
+
+### 2. HTTP Server (`http_server.rb`)
+A standalone HTTP server built with Rack that implements the MCP StreamableHTTP transport protocol. This demonstrates how to create a web-based MCP server with session management and Server-Sent Events (SSE) support.
```
atesgoral:
```suggestion
A standalone HTTP server built with Rack that implements the MCP Streamable HTTP transport. This demonstrates how to create a web-based MCP server with session management and Server-Sent Events (SSE) support.
```

examples/README.md
```diff
@@ -0,0 +1,174 @@
...
+  -H "Content-Type: application/json" \
+  -H "Mcp-Session-Id: $SESSION_ID" \
+  -d '{"jsonrpc":"2.0","method":"tools/call","id":2,"params":{"name":"notification_tool","arguments":{"message":"Hello from cURL!"}}}'
+```
+
+## StreamableHTTP Transport Details
+
+### Protocol Flow
+
+The HTTP server implements the MCP StreamableHTTP transport protocol:
```
atesgoral:
```suggestion
The HTTP server implements the MCP Streamable HTTP transport:
```

lib/mcp/server.rb
```diff
@@ -91,6 +94,14 @@ def define_prompt(name: nil, description: nil, arguments: [], &block)
       @prompts[prompt.name_value] = prompt
     end
 
+    def handle_rack_request(request)
+      @transport.handle_request(request)
+    end
```
kfischer-okarin: I think it would be better if the server interface would not contain methods that only make sense for a certain family of transports..

kfischer-okarin (commented):
I think it might be easier to develop and review if updating the server and transport interfaces to support notifications - including implementing the necessary methods for STDIO was done as a first separate step/PR

Then adding another Transport implementation + examples for Streamable HTTP would be easier to do and review...

Also about the current approach - I think this should be better marked as a `StandaloneHTTPTransport` - since I'm not sure that could be integrated into a Rails/Sinatra/Hanami/etc app as is..... 

But I think the proper road ahead for those frameworks would be maybe to offer separate `mcp-rails-transport` `mcp-sinatra-transport` etc addon gems which provide their specific Transport implementation and the respective frameworks idiomatic Plugin boilerplate

kfischer-okarin:
I also thought a bit about a possible interface... and my idea so far was

1. add a method like this to the `Server`

  ```rb
  def support_notifications(transport, tools_list_changed: false, resources_list_changed: false, resources_subscribe: false, prompts_list_changed: false, logging: false)
  ```
  
  which would be called by the server implementer depending on the needs/capabilities of the application
  
2. Add a `send_notification` or maybe more generic `send_to_client` to the transport (since I think all transports use the same mechanism for notifications and server send requests) - which in case of STDIO would just send to stdout

3. Add all protocol methods as public methods to the server which uses the transport internally to just send it
  ```rb
  def notify_tools_list_changed
    @transport.send_to_client(method: Methods::NOTIFICATIONS_TOOLS_LIST_CHANGED)
  end
  
  def notify_resources_list_changed
    # ...
  
  def notify_resource_changed(uri)
    # ...
  
  def notify_prompts_list_changed
    # ...
  
  def log_to_client(data, level:, logger: nil)
    # ...
  ```
  
  Sending those notifications will be the server implementer's responsibility anyways so adding them easy to use to the server interface seems best....
  
  
I was planning to send a PoC Pull Request sooner or later when I got time... but if you like I could make one quite soon (maybe during the weekend) so it's maybe easier to talk about

lib/mcp/server.rb
Ginja: I can see that POV. No strong opinions over here. I don't think it looks particularly ugly to be able to do this:

```
response = server.handle_rack_request(request)
```

However, if there's consensus I can drop this method so that the above would look like this

```
response = transport.handle_request(request)
``` 

Similar to how `stdio` uses `transport.open`.

Or we modify the current `handle` method in the server that call the right method based on the transport.

```ruby
    def handle(request)
      case transport_type
      when :stdio
        JsonRpcHandler.handle(request) do |method|
          handle_request(request, method)
        end
      when :streamablehttp
        @transport.handle_request(request)
      else
        raise "Unknown transport type: #{transport_type}"
      end
    end
```

Then everything becomes

```
response = server.handle(request)
```

Open to other ideas/suggestions as well 😄

Ginja:
I like what you've proposed for the notifications. I can look to add that to my PR next week.

> Also about the current approach - I think this should be better marked as a StandaloneHTTPTransport - since I'm not sure that could be integrated into a Rails/Sinatra/Hanami/etc app as is.....

I prefer not to rename this. This is providing a barebone implementation of the StreamableHTTP transport. I think it's best to leave it up to implementers as to how they want to expose their MCP servers. Either directly with Rack & a web server (e.g. Puma, Unicorn, WEBrick, etc...) or contained within a web framework (e.g. Rails, Sinatra, etc...).

I prefer to keep this gem free of strong opinions (e.g. no nicities provided for wrapping this in a Rails controller, or including default/expected routes).

> But I think the proper road ahead for those frameworks would be maybe to offer separate mcp-rails-transport mcp-sinatra-transport etc addon gems which provide their specific Transport implementation and the respective frameworks idiomatic Plugin boilerplate

💯 this. It should not be this gem's concern to facilitate the smooth integration into web frameworks. We can however, provide examples in the documentation (which could refer to the potential framework-specific gems).

mcp.gemspec
```diff
@@ -28,6 +28,8 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency("json_rpc_handler", "~> 0.1")
+  spec.add_dependency("rack", ">= 2.0.0")
```
topherbullock:
```suggestion
  spec.add_development_dependency("rack", ">= 2.0.0")
```

`rack` is only used in examples, so it can be marked as a dev dependency.

I want to be mindful of any users who might be using just the stdio transport who won't need rack as a dependency.

lib/mcp/server.rb
topherbullock: `Server` should be kept as unaware of `Transport` concerns as possible.

What concerns me more than the intro of `handle_rack_request` to server is the indirection that this introduces between `Server` and `Transport` 

`Server` having a `transport` attribute and `Transport` having a `server` attribute feels like a step in the wrong direction and leads to this 🔁

https://github.com/modelcontextprotocol/ruby-sdk/blob/78061f788a896909613894b0a541a93a38c70b66/examples/sse_test_server.rb#L67 

Ideally we should keep `handle_request` isolated to the  transport so that a rack-specific transport's `handle_request` is effectively "handle rack request"

All of this behaviour can be removed and the example can be updated to handle requests on the transport.

kfischer-okarin: > Server having a transport attribute and Transport having a server attribute feels like a step in the wrong direction and leads to this 🔁

I see that and am bothered by that too - even after proposing my [interface suggestion](https://github.com/modelcontextprotocol/ruby-sdk/pull/33#issuecomment-2923945335) - and I'm not sure I have a good clean alternative yet..... 🤔 (Maybe I'm overlooking an obvious alternative)

My design is based on the assumption that it would be best to add those `notify_...` methods (which by their event driven nature should be able to be triggered from outside the normal request/response cycle started by `Transport#open`) to the Server interface, since it's the one object knowing about the actual Protocol right now... and keeping that knowledge/concern inside one class seemed like a cleaner design.

But of course, one could depart from that assumption and instead provide a second class for that job - like a `ServerMessageSender` which has a reference to the `Transport` internally to do its job...

But even if we do that - I think eventually the same issue would occur again once the SDK tries to support Sampling... at which point the tool that needs sampling to do its job will need a reference to something that knows the `Transport` in order to send the sampling request to the client - which is the same loop again.........

Ginja:
Added 3 new commits:

* Converted `rack` to a development dependency (00fb685c0007f3a5906a0f37503b0eb3369e9a59)
* Removed transport-specific code from server (44a35b2cacc0e1b2df8584fb4dfa1d1999ae47d1)
* Added initial notification support based on @kfischer-okarin's suggestions (fb2b06b522b11565d1bd8f8a4b33e56e0727eac6)

@kfischer-okarin I added you as a co-author, I hope you don't mind. I figured it was the right thing to do as you proposed the design ❤️ If you prefer I can drop this commit and this support can be added in a dedicated PR. One method I skipped adding was this:

```ruby
def support_notifications(transport, tools_list_changed: false, resources_list_changed: false, resources_subscribe: false, prompts_list_changed: false, logging: false)
```

I wasn't completely sure what the implementation looked like. I figured this can be added later.

Ginja:
Latest set of commits are just a rebase after some new conflicts.

```
Started with run options --seed 28896

  132/132: [===========================================================================================================================================] 100% Time: 00:00:01, Time: 00:00:01

Finished in 1.22367s
132 tests, 405 assertions, 0 failures, 0 errors, 0 skips
```

```
$ bundle exec rubocop
Inspecting 49 files
.................................................

49 files inspected, no offenses detected
```

kfischer-okarin:
@Ginja 

> I added you as a co-author, I hope you don't mind. I figured it was the right thing to do as you proposed the design ❤️ 

Thanks very much appreciated :)

> One method I skipped adding was this:
> ```rb
> def support_notifications(transport, tools_list_changed: false, resources_list_changed: false, resources_subscribe: false, prompts_list_changed: false, logging: false)
> ```

That method is basicallt an explicit expression of the fact that notifications are optional, and since the Object doesn't need a Transport reference unless it supports notifications, I thought it would be ok if we only add it here and not in the constructor.

As for the flags - those would play into determining the capabilities hash

I submitted a little Capabilities related refactor as #42 - if that is adopted.... adding the respective capabilities dynamically depending on which flag is specified should be easy along the lines of
```rb
@capabilities.support_tools_list_changed if tools_list_changed
#...
```

mcp.gemspec
```diff
@@ -29,5 +29,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency("json_rpc_handler", "~> 0.1")
   spec.add_development_dependency("activesupport")
+  spec.add_development_dependency("puma", ">= 5.0.0")
+  spec.add_development_dependency("rack", ">= 2.0.0")
```
topherbullock, line 33: running `ruby examples/http_server.rb` requires adding `rackup` as a dev dependency

```suggestion
  spec.add_development_dependency("rack", ">= 2.0.0")
  spec.add_development_dependency("rackup", ">= 2.1.0")
```

examples/http_server.rb
```diff
@@ -0,0 +1,170 @@
...
+      parsed_body = JSON.parse(body)
+      mcp_logger.info("Request: #{parsed_body["method"]} (id: #{parsed_body["id"]})")
+      mcp_logger.debug("Request body: #{JSON.pretty_generate(parsed_body)}")
+    rescue JSON::ParserError
+      mcp_logger.warn("Request body (raw): #{body}")
+    end
+  end
+
+  # Handle the request
+  puts "request: #{request.inspect}"
```
topherbullock:
```suggestion
```

this logs the whole `Rack::Request`, which pollutes the logs

lib/mcp/server/transports/streamable_http_transport.rb
```diff
@@ -0,0 +1,295 @@
...
+        def send_ping_to_stream(stream)
+          stream.write(": ping #{Time.now.iso8601}\n\n")
+          stream.flush if stream.respond_to?(:flush)
+        end
+
+        def handle_post(request)
+          body_string = request.body.read
+          session_id = extract_session_id(request)
+
+          body = parse_request_body(body_string)
```
topherbullock, line 100: `body_string` can be `nil` here, `POST` requests with an empty request body ( from testing via the MCP inspector ) raised in the http example:

```
TypeError: no implicit conversion of nil into String (TypeError)

    Parser.parse(source, opts)
                 ^^^^^^^^^^^^
	/opt/homebrew/lib/ruby/gems/3.4.0/gems/json-2.12.2/lib/json/common.rb:338:in 'JSON::Ext::Parser.parse'
	/opt/homebrew/lib/ruby/gems/3.4.0/gems/json-2.12.2/lib/json/common.rb:338:in 'JSON.parse'
	examples/http_server.rb:131:in 'block in <main>'
```

`parse_request_body` tries `JSON.parse(nil)` and throws an exception;  this should be handled as an invalid request similar to invalid json string

lib/mcp/server/transports/streamable_http_transport.rb
```diff
@@ -0,0 +1,295 @@
...
+          @sessions.delete(session_id)
+        end
+
+        def extract_session_id(request)
+          request.env["HTTP_MCP_SESSION_ID"]
+        end
+
+        def parse_request_body(body_string)
+          JSON.parse(body_string)
+        rescue JSON::ParserError
```
topherbullock:
```suggestion
        rescue JSON::ParserError, TypeError
```
rescue on `TypeError` would address issues with nil request bodies.

examples/http_server.rb
Ginja: 👍  This was to showcase what was being sent to help debug. No strong opinions on keeping it though.

mcp.gemspec
Ginja: Good catch!

lib/mcp/server/transports/streamable_http_transport.rb
Ginja: Great catch!

examples/http_server.rb
```diff
@@ -0,0 +1,169 @@
...
+      mcp_logger.warn("Request body (raw): #{body}")
+    end
+  end
+
+  # Handle the request
+  response = transport.handle_request(request)
+
+  # Log the MCP response details
+  _, _, body = response
+  if body.is_a?(Array) && !body.empty?
```
topherbullock:
```suggestion
  if body.is_a?(Array) && !body.empty? && body.first
```

I'm still seeing some errors when testing this in the MCP Inspector, but the server still works past this error.


```
[MCP] Request body: {
  "jsonrpc": "2.0",
  "method": "notifications/initialized"
}
TypeError: no implicit conversion of nil into String (TypeError)

    Parser.parse(source, opts)
                 ^^^^^^^^^^^^
	/Users/topherbullock/.gem/ruby/3.3.0/gems/json-2.12.0/lib/json/common.rb:338:in `parse'
	/Users/topherbullock/.gem/ruby/3.3.0/gems/json-2.12.0/lib/json/common.rb:338:in `parse'
	./examples/http_server.rb:130:in `block in <main>'
```

topherbullock (approved):
thanks for this @Ginja!

README.md
```diff
@@ -45,20 +47,53 @@ It implements the Model Context Protocol specification, handling model context r
...
+#### Notification Methods
+
+The server provides three notification methods:
+- `notify_tools_list_changed()` - Send a notification when the tools list changes
+- `notify_prompts_list_changed()` - Send a notification when the prompts list changes  
+- `notify_resources_list_changed()` - Send a notification when the resources list changes
+
+#### Notification Format
+
+Notifications follow the JSON-RPC 2.0 specification and use these method names:
```
kfischer-okarin, line 63: This section seems unnecessarily specific and redundant about JSON-RPC - this is mentioned in the main section already

examples/http_server.rb
```diff
@@ -0,0 +1,169 @@
...
+
+# Start the server
+puts "Starting MCP HTTP server on http://localhost:9292"
+puts "Use POST requests to initialize and send JSON-RPC commands"
+puts "Example initialization:"
+puts '  curl -X POST http://localhost:9292 -H "Content-Type: application/json" -d \'{"jsonrpc":"2.0","method":"initialize","id":1,"params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1.0"}}}\''
+puts ""
+puts "The server will return a session ID in the Mcp-Session-Id header."
+puts "Use this session ID for subsequent requests."
+puts ""
```
kfischer-okarin, line 164: This mentions the Mcp-Session-Id header and it is somewhat (I think not correctly though) supported in the client - but the server does not actually create or emit any such headers...

examples/sse_test_server.rb
```diff
@@ -0,0 +1,173 @@
...
+
+# Build the Rack application with middleware
+rack_app = Rack::Builder.new do
+  use(Rack::CommonLogger, Logger.new($stdout))
+  use(Rack::ShowExceptions)
+  run(app)
+end
+
+# Print usage instructions
+puts "=== MCP SSE Test Server ==="
```
kfischer-okarin: I think this example name is misleading since it can make you easily think this is about the legacy SSE transport - it should be better called something like Streamable HTTP with Notificaitons etc

examples/sse_test_server.rb
```diff
@@ -0,0 +1,173 @@
...
+puts "2. echo - Simple echo tool"
+puts ""
+puts "Testing SSE:"
+puts ""
+puts "1. Initialize session:"
+puts '   curl -X POST http://localhost:9393 -H "Content-Type: application/json" \\'
+puts '     -d \'{"jsonrpc":"2.0","method":"initialize","id":1,"params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"sse-test","version":"1.0"}}}\''
+puts ""
+puts "2. Connect SSE stream (use the session ID from step 1):"
+puts '   curl -N -H "Mcp-Session-Id: YOUR_SESSION_ID" http://localhost:9393'
```
kfischer-okarin: Same issue as with the other example... mcp-session-id is not actually emitted by the server

lib/mcp/transport.rb
```diff
@@ -2,32 +2,57 @@
...
+        @transports ||= {}
+        @transports[name] = transport_class
+      end
+
+      def get(name)
+        raise "Transport #{name} not found" unless @transports
+
+        @transports.fetch(name) { raise "Transport #{name} not found" }
+      end
+    end
```
kfischer-okarin: What is this Transport Registration mechanism for? I'm not sure it's necessary - or at least not for this particular PR

test/mcp/server/transports/stdio_notification_integration_test.rb
```diff
@@ -0,0 +1,249 @@
...
+
+require "test_helper"
+require "json"
+require "mcp/server"
+require "mcp/server/transports/stdio_transport"
+
+module MCP
+  class Server
+    module Transports
+      class StdioNotificationIntegrationTest < ActiveSupport::TestCase
```
kfischer-okarin, line 11: I think integration tests are something worth doing but not sure this is the best approach - going one level higher, i.e. actually starting an example as a process and really having something like a recorded interaction script - sending stdin/stdout might be more effective

Again also questionable whether it should be handled in this same PR

test/mcp/server_notification_test.rb
```diff
@@ -0,0 +1,136 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+module MCP
+  class ServerNotificationTest < ActiveSupport::TestCase
```
kfischer-okarin, line 7: I wonder why this test is separate - I think adding pinpoint new assertions to the existing ServerTest might be better - there's also some odd tests in here like checking constants are defined 🤔

kfischer-okarin (changes requested):
Sorry for not reviewing and commenting on this earlier - I was mainly concerned with the Server class and Transport changes and did not look at tests or the examples.

But in my opinion the both tests and examples still have some problems which I would prefer not to be merged (but of course I can understand if the main maintainers want to move forward and fix these issues later - so feel free to dismiss as just one contributors opinion)

I gave a few representative examples of the issues I personally see

No offense intended, but my general impression while reading was that examples and tests have maybe been generated by an coding agent (which I totally support) but not deeply reviewed
(this again is of course something which the main maintainer team might have a different opinion on - i.e. it's fine to merge code like that because with Coding Agents it's also as easy to change it again later... I personally don't think that is a good approach but of course respect the maintainers direction)

kfischer-okarin:
Also as an additional note:

I'm half-way finished working on a little minimal version containing a subset of this branch that implements the general  list changed notifications support for STDIO - (of course properly attributing @Ginja as Co-Author everywhere where I used any code from this PR) - since this PR seems to get quite big and if the discussion/review goes on it might still take a while until any value is provided to the main branch...

Unless you just want to go on with this branch as is which is also fine of course ;)

examples/http_server.rb
Ginja: > but the server does not actually create or emit any such headers...

Unless I'm misunderstanding, I believe it does?

```
$ curl -D - -s -o /dev/null -X POST http://localhost:9292/mcp \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","method":"initialize","id":1,"params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"curl-test","version":"1.0"}}}'
HTTP/1.1 200 OK
Content-Type: application/json
Mcp-Session-Id: e286fd80-59b8-4281-9bdd-39e88d4d2f26
Content-Length: 184
```

examples/sse_test_server.rb
Ginja: Nit, but SSE is a supported transport/mechanism even within Streamable HTTP. Also as defined [here](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports):

> In the Streamable HTTP transport, the server operates as an independent process that can handle multiple client connections. This transport uses HTTP POST and GET requests. Server can optionally make use of [Server-Sent Events](https://en.wikipedia.org/wiki/Server-sent_events) (SSE) to stream multiple server messages.

> It should be better called something like Streamable HTTP with Notificaitons etc

This test is showcasing the SSE mechanism. I'm fine leaving it as-is, and fixing in another PR if someone feels strongly about it.

examples/sse_test_server.rb
Ginja: It is. See my example above.

lib/mcp/transport.rb
Ginja: Convenience method in case it's of use to any implementors. I originally was working off some commits from @topherbullock, and I believe this was in there and so I kept it.

test/mcp/server_notification_test.rb
Ginja: I think it's worth checking if the constants are defined as expected as the values are currently what the spec is defining. If someone inadvertently changes them problems could arise.

lib/mcp/transport.rb
topherbullock: I think we can remove this until it is actually used anywhere

examples/sse_test_server.rb
topherbullock: `streaming_http_test_server.rb` would be a good alternative name to reflect this isn't related to SSE. Agree it could be somewhat confusing at a glance.

test/mcp/server_notification_test.rb
```diff
@@ -0,0 +1,136 @@
...
+      assert_equal Methods::NOTIFICATIONS_TOOLS_LIST_CHANGED, notifications[0][:method]
+      assert_equal Methods::NOTIFICATIONS_PROMPTS_LIST_CHANGED, notifications[1][:method]
+      assert_equal Methods::NOTIFICATIONS_RESOURCES_LIST_CHANGED, notifications[2][:method]
+    end
+
+    test "notification method constants are defined correctly" do
+      assert_equal "notifications/tools/list_changed", Methods::NOTIFICATIONS_TOOLS_LIST_CHANGED
+      assert_equal "notifications/prompts/list_changed", Methods::NOTIFICATIONS_PROMPTS_LIST_CHANGED
+      assert_equal "notifications/resources/list_changed", Methods::NOTIFICATIONS_RESOURCES_LIST_CHANGED
+    end
```
topherbullock:
```suggestion
```

test/mcp/server_notification_test.rb
topherbullock: I'd prefer to rely on functional tests that rely on the constant values rather than the existing assertions on constant values. Other tests should break if a method constant is removed.

test/mcp/server/transports/stdio_notification_integration_test.rb
topherbullock: I'm fine with this as is, since there's currently no integration tests for the correctness of transports. 

I agree a test that starts a process and sends stdin / stdout would be ideal for stdio integration, but that is a fix that can be applied forward after this merges

examples/http_server.rb
kfischer-okarin: I'm sorry 🙈  it seems I missed checking the Transport Implementation itself - I only looked in the respective example servers. Yes it's implemented there 👍🏻

test/mcp/transport_test.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "mcp/server"
+require "mcp/transport"
+
+module MCP
+  class TransportTest < ActiveSupport::TestCase
```
kfischer-okarin, line 8: Not sure these tests add any value anymore after the registration mechanism was removed

lib/mcp/transport.rb
```diff
@@ -2,32 +2,44 @@
...
 
-    def handle_json_request(request)
-      response = @server.handle_json(request)
-      send_response(response) if response
+    # Send a notification to the client
+    # Returns true if the notification was sent successfully
+    def send_notification(method, params = nil)
+      raise NotImplementedError, "Subclasses must implement send_notification"
     end
   end
```
kfischer-okarin, line 44: This is a rather fuzzy big picture design feedback - but I was wondering what to include in the `Transport` interface and what not...

The main goal of the Transport interface from my point of view is **hiding the details of the transport method from the client code**. 

## Clients of the Transport object so far
The main client until now was the main application entrypoint and the only really essential public interface methods were `open` and `close`. Everything else can be easily handled inside the loop created by `open` - so technically there is no need for any of the other methods to be public.

Now with the current addition of the notify methods the server becomes an additional client. And the only thing the server needs to know is how to send a notification via the transport. So `send_notification` is another must-have addition to the interface.

## Issues surfaced through adding the Streamable HTTP implementation

### What about the Rack Setup?
If we wanted to keep the pure simple `open` interface - it would be necessary to somehow include all the necessary Rack Setup in the current Transport class.

But considering the wider point of view that the natural Rack interface would be just exposing a callable taking a rack env object - maybe the `open` interface isn't the best way to handle things after all?

## Make the Transport interface super simple?
So maybe removing `open` (and accordingly `close` I guess) from the interface and just have each Transport document their own preferred way of starting itself might be another valid way to do things - so the only method included in the public interface is really `send_notificaton`

But even in that case I'd still personally say that it would be nicer to make the Streamable HTTP a bit more out-of-the back Rack startable than it is now... which probably just means:
- exposing a way to add middlewares
- adding an option for the endpoint name (/mcp by default)
- ...?

test/mcp/transport_test.rb
topherbullock: this test covers the shared `Transport` functionality.

lib/mcp/transport.rb
topherbullock: > make the Streamable HTTP a bit more out-of-the back Rack startable than it is now

yeah, I think the next direction after this merges should be to allow base `StreamableHTTP` transport to be easily dropped into any Rack, Rails, etc. app - build out conventions that make adding to any existing server trivial.

this gives us a base we can work from and a reference implementation, so I'm keen to get this merged and iterate forward

topherbullock (approved):
thanks for the clean up @Ginja. This looks good to merge. 
gives us a reference for future improvements that make adding http transport to any rack/rails app trivial.

we can fix forward with any tweaks / extensions

lib/mcp/transport.rb
topherbullock: @kfischer-okarin let's continue the drop-in rack chat here #59.
I've made a separate one for Rails #60

kfischer-okarin (approved):
Yes I guess merging it as is and then continuing improvements in further smaller PRs is maybe the better way to go otherwise this PR will be a blocker too long 👍🏻

Animeshz:
Documentation missing?

PR #42, opened 2025-06-04
Refactor: Extract Capabilities object

kfischer-okarin:
I extracted a Capabilities object, which allows setting the capabilities from the server incrementally and encapsulates the actual calculation of the final capabilities hash.

## Motivation and Context
The current `determine_capabilities` approach would need to become a bit more complex to support all the sub-capabilities without interfering with the already supported auto-determination. This would need a bit of custom bookkeeping which if kept in the server would cause quite a bit of cognitive noise. So I extracted it to its own object with an idempotent "support_..." API.

This is originally thought as a refactor to prepare for easy implementation of the `Server#support_notifications` method [I proposed in #33](https://github.com/modelcontextprotocol/ruby-sdk/pull/33#issuecomment-2923945335) - but even if that API should not be adopted after all, I still think it's nice to hide all that capability determination logic in its own object.

## How Has This Been Tested?
I did not test this manually, but the existing unit tests still pass.

## Breaking Changes
No

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
None

### DIFF
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -46,6 +46,10 @@ module MCP
       @resource_index = index_resources_by_uri(resources)
       @server_context = server_context
       @configuration = MCP.configuration.merge(configuration)
+      @capabilities = Capabilities.new(capabilities)
+      @capabilities.support_tools if tools.any?
+      @capabilities.support_prompts if prompts.any?
+      @capabilities.support_resources if resources.any? || resource_templates.any?
 
       @handlers = {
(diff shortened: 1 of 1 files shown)


kfischer-okarin:
BTW I had some help from Claude Code (as pointed out in the commit, too), especially with writing out all the Unit Tests - feel free to delete some of them if you find them too extensive

atesgoral (approved):

PR #43, opened 2025-06-04
Validate ToolCall Args against Schema

tylerrowsell:
## Motivation and Context
Tool call arguments are not being validated against the schema. This ensures that the schema is valid and that the tool call args are valid. Otherwise it fails early. 

There is an option to disable tool call validation.

## How Has This Been Tested?
Yes. Tested on internal products.

## Breaking Changes
Tool calls will be validated, if your application has tool calls coming in that don't match the schema these requests will now fail.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -150,7 +150,7 @@ module MCP
           report_exception(e, { request: request })
           if e.is_a?(RequestHandlerError)
             add_instrumentation_data(error: e.error_type)
-            raise e
+            raise StandardError, e.message
           end
 
           add_instrumentation_data(error: :internal_error)
@@ -211,6 +211,15 @@ module MCP
           request,
           error_type: :missing_required_arguments,
         )
+      end
+
+      if configuration.validate_tool_call_arguments && tool.input_schema
+        begin
+          tool.input_schema.validate_arguments(arguments)
+        rescue Tool::InputSchema::ValidationError => e
+          add_instrumentation_data(error: :invalid_schema)
+          raise RequestHandlerError.new(e.message, request, error_type: :invalid_schema)
+        end
       end
 
       begin
--- a/lib/mcp/tool/input_schema.rb
+++ b/lib/mcp/tool/input_schema.rb
@@ -1,13 +1,18 @@
 # frozen_string_literal: true
 
+require "json-schema"
+
 module MCP
   class Tool
     class InputSchema
+      class ValidationError < StandardError; end
+
       attr_reader :properties, :required
 
       def initialize(properties: {}, required: [])
         @properties = properties
         @required = required.map(&:to_sym)
+        validate_schema!
       end
 
       def to_h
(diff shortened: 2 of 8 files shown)


mcp.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency("json_rpc_handler", "~> 0.1")
+  spec.add_dependency("json-schema", "~> 4.1")
```
kfischer-okarin, line 31: Why did you use json-schema over [json_schemer](https://github.com/davishmcclurg/json_schemer)

topherbullock: [json-schema](https://rubygems.org/gems/json-schema) has fewer runtime dependencies, and has broader adoption compared to [json_schemer](https://rubygems.org/gems/json_schemer).   are there functional benefits of using `json_schemer` instead?

topherbullock (dismissed):

mcp.gemspec
kfischer-okarin: > are there functional benefits of using json_schemer instead?

No not really I think - I was just putting it out there as a question - since I found myself reaching json_schemer in my own side projects

But the older json-schema gem seems to be the more stable simpler choice in this case (for a library) - and tool schemas a very basic so no need for any new features provided by newer JSON Schema drafts

👍🏻

lib/mcp/tool/input_schema.rb
```diff
@@ -21,6 +26,32 @@ def missing_required_arguments?(arguments)
...
+            accept_uri: false,
+            accept_file: ->(path) { path.to_s.start_with?(Gem.loaded_specs["json-schema"].full_gem_path) },
+          )
+          metaschema = JSON::Validator.validator_for_name("draft4").metaschema
+          errors = JSON::Validator.fully_validate(metaschema, schema, schema_reader: schema_reader)
+          if errors.any?
+            raise ArgumentError, "Invalid JSON Schema: #{errors.join(", ")}"
+          end
+        rescue StandardError => e
+          raise ArgumentError, "Invalid JSON Schema: #{e.message}"
```
koic: This exception handling looked a bit odd to me, as it uses a generic exception class like `StandardError` to represent a specific error such as "Invalid JSON Schema". It might be more appropriate either to let the exception propagate to the caller without handling it here, or to replace `StandardError` with a more specific exception class.

tylerrowsell: Good point. Resolved in most recent commit.

lib/mcp/server.rb
```diff
@@ -150,7 +150,7 @@ def handle_request(request, method)
           report_exception(e, { request: request })
           if e.is_a?(RequestHandlerError)
             add_instrumentation_data(error: e.error_type)
-            raise e
+            raise StandardError, e.message
```
koic: Is it necessary to replace it with `StandardError`?

topherbullock (approved):

Issue #39, opened 2025-06-03
Unintuitive use of meta programming for :server context in tool call


sergioisidoro:
I've spent a good deal of time trying to figure out why `:server_context` was not being passed into the tool call
I was using `*args` and `**kargs` to debug things and still could not figure out why I could not get the server context in tool calls. I was going crazy.

Finally went looking at the code, and it seems to do some reflection and change its behaviour in a bit strange ways.
```
        call_params = tool_call_parameters(tool)

        if call_params.include?(:server_context)
          tool.call(**arguments.transform_keys(&:to_sym), server_context:).to_h
        else
          tool.call(**arguments.transform_keys(&:to_sym)).to_h
        end
```

Since context is used for authentication and authorization this is a bit strange to me. Why not always pass :server_context to a tool call and introduce complexity with this introspection, that only works if the signature explicitly mentions the parameter? 🤔

kfischer-okarin:
It should probably be better documented, but I'd personally prefer if the server_context is not passed as a useless argument if I don't need to use it in the tool... 

Local running MCP servers don't need this property since the tools can just read the Environment Variables or local config files directly.

sergioisidoro:
That's fair, but I think the method of reflection is brittle and breaks a few common patterns of ruby (eg. *args **kwargs). Also something to have in consideration would be to use defensive programing, ensuring that the tool call cannot in any instance pass the argument `server_contex` as a parameter to the tool (possibly overriding parameters used for permission checks)

Using positional arguments or keyword arguments would be a defensive approach. ie `call(:tool_params, :server_context)`

And even in local MCP servers, relying on the environment in the tool call does not sound super appropriate, as you may want to have shared context across all tools. Calling environment in the tool call would have a significant code replication across all tools

sergioisidoro:
Quick update: It seems that everything works as expected in the tool call in the tests of this repo, even when I add more extensive tests with `*args, **kwargs`, and other variations.

I'm a bit lost here as to why `server_context` is not passed on some occasions. My best guess is that my environment `tool_call_method_def` does not return the same results reliably. If I remove the if condition mentioned above from the gem in my bundle, everything works as expected.

Running this on `ruby:3.3.5` without sorbet

---
Edit:
The only way to get this to work with with splats, is by doing this
```.rb            
    def self.call(*args, server_context: nil, **tool_arguments)
```

Which is unpleasant, to say the least.

kfischer-okarin:
[...]

PR #54, opened 2025-06-09
Make `server_context` optional for Tools and Prompts

julianojulio:
This PR makes the `server_context` parameter truly optional for both tools and prompts, improving type checker compatibility and addressing developer experience concerns.

## Motivation and Context

The previous implementation required `server_context` as a keyword argument in the base `Tool` and `Prompt` classes, which caused several issues:

1. **Type checking conflicts**: Tools and prompts that didn't need server context were forced to accept it, creating type signature mismatches when using type checkers like Sorbet
2. **Inflexible tool/prompt definitions**: Every tool and prompt had to declare `server_context:` in its method signature, even when unused
3. **Unintuitive parameter handling**: As reported in issue #39, the implementation used meta-programming patterns that were difficult to understand and predict

## How Has This Been Tested?

Added comprehensive test coverage in `test/mcp/server_context_test.rb` covering all scenarios:

**For Tools:**
- Tools without server_context parameter
- Tools with optional server_context  
- Tools with required server_context
- Tools using `**kwargs` for flexible parameter handling

**For Prompts:**
- Prompts without server_context parameter
- Prompts with optional server_context
- Prompts with required server_context  
- Prompts with flexible parameter signatures

All existing tests have been updated and continue to pass.

## Breaking Changes

**No breaking changes.** This change maintains full backward compatibility:
- Tools and prompts with explicit `server_context` parameters continue to work unchanged
- Tools and prompts can now omit the parameter entirely if they don't need it
- The server intelligently detects what each tool/prompt expects

### 1. Made server_context optional in base classes

```ruby
# In Tool class
def call(*args, server_context: nil)
  # ...
end

# In Prompt class  
def template(args, server_context: nil)
  # ...
end
```

### 2. Improved parameter detection

The new implementation properly inspects method parameters to determine if a tool/prompt accepts server_context:

```ruby
def call_tool_with_args(tool, arguments)
  args = arguments.transform_keys(&:to_sym)
  
  # Check if the tool accepts server_context or has **kwargs
  parameters = tool.method(:call).parameters
  accepts_server_context = parameters.any? { |_type, name| name == :server_context }
  has_kwargs = parameters.any? { |type, _| type == :keyrest }
  
  if accepts_server_context || has_kwargs
    tool.call(**args, server_context: server_context).to_h
  else
    tool.call(**args).to_h
  end
end
```

This approach:
- Preserves parameter type information for accurate detection
- Handles tools/prompts with explicit `server_context` parameters
- Supports tools/prompts using `**kwargs` for maximum flexibility
- Makes behavior predictable and transparent

### 3. Usage Examples

Now you can define tools and prompts more flexibly:

```ruby
# Simple tool - no server_context needed
class SimpleTool < Tool
  def self.call(message:)
    Tool::Response.new([{ type: "text", content: message }])
  end
end

# Optional context tool
class OptionalContextTool < Tool  
  def self.call(message:, server_context: nil)
    user = server_context&.dig(:user) || "anonymous"
    Tool::Response.new([{ type: "text", content: "#{message} from #{user}" }])
  end
end

# Flexible tool with kwargs
class FlexibleTool < Tool
  def self.call(**kwargs)
    # Can access both regular args and server_context if provided
    Tool::Response.new([{ type: "text", content: "Received: #{kwargs.inspect}" }])
  end
end
```

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This PR addresses [issue #39: Unintuitive use of meta programming for :server context in tool call](https://github.com/modelcontextprotocol/ruby-sdk/issues/39) by making the server_context handling explicit and transparent.

The change improves compatibility with Ruby type checkers (like Sorbet) and makes the gem more accessible to teams using static typing, while maintaining full backward compatibility for existing implementations.


lib/mcp/server.rb
```diff
@@ -292,5 +286,33 @@ def tool_call_method_def(tool)
...
+      parameters = prompt.method(:template).parameters
+      accepts_server_context = parameters.any? { |_type, name| name == :server_context }
+      has_kwargs = parameters.any? { |type, _| type == :keyrest }
+
+      if accepts_server_context || has_kwargs
+        prompt.template(args, server_context: server_context).to_h
+      else
+        prompt.template(args).to_h
+      end
+    end
```
kfischer-okarin: I think this could be simplified into a single function like
```rb
def accepts_server_context?(method_object)
  parameters = method_object.parameters
  # ...
  accepts_server_context || has_kwargs
end
```

To have the signature checking logic be cleanly contained in a single place.

kfischer-okarin: Also independent of that an old function `tool_call_parameters` still exists which becomes unused by the current version of the change. So either deleting or adjusting that to include these changes would be good.

sambostock (commented):
#56 explores the idea of making tools & prompts into simple data/config/value objects, and gets rid of the inheritance based approach in favour of the `define` based approach.

One of the benefits of this is that the implementation of the tool/prompt _always_ uses a block, and blocks implicitly handle omitting arguments, meaning we can get rid of all the logic to do with conditionally passing `server_context:`.

julianojulio:
Hey @kfischer-okarin, anything else to add here?

kfischer-okarin (approved):
Depending on how #56 turns out this might needs to change or even become unnecessage but this PR as it is now looks good to me for approve for now

topherbullock (approved):

PR #55, opened 2025-06-10
Test example code to ensure it runs

sambostock:
This PR introduces test files which run the code snippets in `/examples` and in `README.md`. As each scenario is unique and requires its own setup and assertions, each test is bespoke (rather than being auto-generated).

## Motivation and Context

As part of #48 and the surrounding exploration of using simpler "definition" classes, I realized I'd have to ensure the example code still worked, so I thought I'd explore the idea of actually testing it.

It turns out this paid off, as a bunch of the examples were either broken themselves, or revealed bugs in the implementation.

## How Has This Been Tested?

The PR consists almost entirely of tests. 😅 

## Breaking Changes

This should not include any breaking changes.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

I have added `<!-- SNIPPET ID: snippet_name -->` comments to the snippets we need to be able to extract from `README.md`. These will be invisible to readers, but make it much simpler to find the snippets than needing to hard code their contents or some other such shenanigans.


test/integration/readme_code_snippets_test.rb
```diff
@@ -0,0 +1,324 @@
...
+    )
+
+    skip "FIXME: this next code snippet is invalid and there doesn't seem to be a way to make both pass..."
+
+    assert_json_lines(
+      [
+        { jsonrpc: "2.0", id: "1", result: { content: [{ type: "text", text: "OK" }], isError: false } },
+      ],
+      run_code_snippet("tool_definition_with_block"),
+    )
```
sambostock, line 160: Currently, this is the relevant snippet.

> ```ruby
> tool = MCP::Tool.define(
>   name: "my_tool",
>   description: "This tool performs specific functionality...",
>   annotations: {
>     title: "My Tool",
>     read_only_hint: true
>   }
> ) do |args, server_context|
>   Tool::Response.new([{ type: "text", text: "OK" }])
> end
> ```

It fails with the following error

```
 FAIL ReadmeCodeSnippetsTest#test_Tools_examples_work_exactly_as_documented_in_README (1.66s)
        Expected no stderr in: Exception: #<MCP::Server::RequestHandlerError: Internal error calling tool my_tool>
        Exception cause: #<ArgumentError: wrong number of arguments (given 1, expected 2)>
        .
        Expected "Exception: #<MCP::Server::RequestHandlerError: Internal error calling tool my_tool>\n" +
        "Exception cause: #<ArgumentError: wrong number of arguments (given 1, expected 2)>\n"
         to be empty.
        test/integration/readme_code_snippets_test.rb:312:in 'block (2 levels) in ReadmeCodeSnippetsTest#run_code_snippet'
        test/integration/readme_code_snippets_test.rb:306:in 'Dir.chdir'
        test/integration/readme_code_snippets_test.rb:306:in 'block in ReadmeCodeSnippetsTest#run_code_snippet'
        /opt/rubies/3.4.4/lib/ruby/3.4.0/tmpdir.rb:105:in 'Dir.mktmpdir'
        test/integration/readme_code_snippets_test.rb:305:in 'ReadmeCodeSnippetsTest#run_code_snippet'
        test/integration/readme_code_snippets_test.rb:154:in 'block in <class:ReadmeCodeSnippetsTest>'
```

I tried simply switching to a `server_context:` keyword block argument, but that also fails

```
 FAIL ReadmeCodeSnippetsTest#test_Tools_examples_work_exactly_as_documented_in_README (2.13s)
        Expected no stderr in: Exception: #<MCP::Server::RequestHandlerError: Internal error calling tool my_tool>
        Exception cause: #<ArgumentError: wrong number of arguments (given 0, expected 1; required keyword: server_context)>
        .
        Expected "Exception: #<MCP::Server::RequestHandlerError: Internal error calling tool my_tool>\n" +
        "Exception cause: #<ArgumentError: wrong number of arguments (given 0, expected 1; required keyword: server_context)>\n"
         to be empty.
        test/integration/readme_code_snippets_test.rb:312:in 'block (2 levels) in ReadmeCodeSnippetsTest#run_code_snippet'
        test/integration/readme_code_snippets_test.rb:306:in 'Dir.chdir'
        test/integration/readme_code_snippets_test.rb:306:in 'block in ReadmeCodeSnippetsTest#run_code_snippet'
        /opt/rubies/3.4.4/lib/ruby/3.4.0/tmpdir.rb:105:in 'Dir.mktmpdir'
        test/integration/readme_code_snippets_test.rb:305:in 'ReadmeCodeSnippetsTest#run_code_snippet'
        test/integration/readme_code_snippets_test.rb:154:in 'block in <class:ReadmeCodeSnippetsTest>'
```

Noticing https://github.com/modelcontextprotocol/ruby-sdk/pull/54, I suspect @julianojulio may have run into something similar.

koic:
I think the concept is solid. In this case, having the commits split makes it harder to follow later, so can you squash them into a single commit?

lib/mcp/server.rb
```diff
@@ -9,6 +9,14 @@ module MCP
   class Server
     DEFAULT_VERSION = "0.1.0"
 
+    class << self
+      def protocol_version = MCP.configuration.protocol_version
+
+      def protocol_version=(version)
+        MCP.configuration.protocol_version = version
+      end
+    end
```
koic: Is it possible that the documentation was actually incorrect in this case?
It's unclear whether this API, which hasn't been provided until now, is truly necessary. It would be better to keep the published APIs carefully curated.

test/fixtures/files/code_snippet_wrappers/readme/instrumentation_callback.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require "mcp"
+
+MCP.configure do |config|
+  eval(File.read("code_snippet.rb"), binding) # rubocop:disable Security/Eval -- We need to run the snippet in this context to pass the config
```
koic: Nits: Excluding it in `.rubocop.yml` would be the simpler approach.

```diff
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,3 +4,7 @@ inherit_gem:
 plugins:
   - rubocop-minitest
   - rubocop-rake
+
+Security/Eval:
+  Exclude:
+    - test/fixtures/files/**/*
```

sambostock:
@koic I had kept the commits granular so it was clear which changes were required by which tests, in particular so reviewers can discuss them separately. I'm happy to squash after addressing review feedback though 👍

test/fixtures/files/code_snippet_wrappers/readme/instrumentation_callback.rb
sambostock: Yeah, good idea! 👍

lib/mcp/server.rb
sambostock: Yeah, I agree. I made the documentation example work, but intended to discuss. I agree that there doesn't seem to be a reason to expose the documented API; the consumer should be able to edit the configuration directly, especially since it allows granularly setting the protocol version per server if required.

sambostock: Incidentally there is `MCP::Server::DEFAULT_VERSION` which we expose in `init` in `server_info`, but is that something we should be exposing to clients? This is effectively the gem version, isn't it? Don't clients only care about the protocol version? Should we be leaking the gem version?

If so, it doesn't seem like something the consumer should actually be able to change, and should probably be hard coded to the gem version. Unless I'm missing something... 🤔

koic: Based on my understanding of the PR #7, `MCP::Server::DEFAULT_VERSION` is defined as a server version that is independent of the gem version `MCP::VERSION`. I think these are separate and distinct values. The more detailed intent might be something @kfischer-okarin can speak to.

sambostock: Ok, I'll leave that untouched in this PR then. 👍

test/integration/readme_code_snippets_test.rb
```diff
@@ -0,0 +1,328 @@
...
+
+    response = {
+      jsonrpc: "2.0",
+      id: "1",
+      error: { code: -32603, message: "Internal error", data: "Internal error calling tool error_tool" },
+    }.to_json
+
+    assert_equal(<<~STDOUT, stdout)
+      Bugsnag notified of #{error.inspect} with metadata #{metadata.inspect}
+      Got instrumentation data #{instrumentation_data}
```
sambostock, line 89: Interestingly, because Ruby changed `Hash#to_s`'s output for `Symbol` keys, we have to interpolate instead of inlining the value (so the test passes across Ruby versions), and because RuboCop detects we're interpolating a literal and inlines it, we have to extract a variable.

sambostock:
@koic I've squashed the commits and addressed your feedback 👍

The only thing remaining is addressing the test I had to `skip`. I'm not sure what we want to do there, or if it's worth fixing, since there are multiple PRs open that might affect it (#54, #56).

sambostock:
I've rebased again, just to keep this in sync with `main`. #33 added some more examples, so we'll have to add the tests for those too, but I think it would make sense to merge this sooner than later, and open a followup PR to add the missing tests.

topherbullock (changes requested):
I'm not really into the `<!-- SNIPPET ID: ... -->` formatting in the README.md, this ties us to our own invented formatting and isn't obvious at a glance.

I'd prefer to focus the project on having good integration testing around everything in `examples/` and ensure the Readme is up to date before adding something like this. 

`test/fixtures/files/code_snippet_wrappers/` already means we have test logic for every block of example code, and I'd prefer having these code blocks in ruby files rather than testing code blocks extracted from the Readme 

Could you pivot this to include all the readme code snippets as files in the `examples/` dir, and test those instead?

koic:
I'll close this PR since it has diverged from the current implementation. The documentation examples have been improved, and as @topherbullock noted, the necessity of a special annotation appears to have decreased. Thank you.

PR #61, opened 2025-06-14
Remove automatic capability determination and add missing capabilities

kfischer-okarin:
This is effectively reverting my two previous PRs #8 #42 and adding the missing capabilities for list changed notifications.

See the Discussion linked in Motivation and context for details.

## Motivation and Context
See Discussion https://github.com/modelcontextprotocol/ruby-sdk/discussions/57 for background

## How Has This Been Tested?
I tried it out in the inspector

## Breaking Changes
Not really - for the same reasoning as already mentioned in #8 
> This should not be a breaking change since an empty tool list and no capability should make no practical difference to a procotol conform MCP client.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
The notification capabilities are not yet checked like the normal method capabilities - but that should be fixed in another PR I think (I will send that too later)


atesgoral:
This makes sense, especially after I went around digging for the root cause of an issue I ran into: I wasn't passing any tools at init, but calling `define_tool` for custom bindings. The capability magic doesn't kick in then. We could either forward fix that issue (should be simple) or go ahead with this explicit method.

@topherbullock @koic what do you think?

atesgoral (approved):

koic:
Looks good to me. Thanks!

PR #64, opened 2025-06-17
Remove duplicate `method` instrumentation

sambostock:
This removes duplicate calls to `add_instrumentation_data(method: ...)`.

## Motivation and Context

`handle_request` already adds the `method` to the instrumentation data, so there is no need to repeat this. Moreover, consumers can override these anyway.

## How Has This Been Tested?

This is covered by existing tests.

## Types of changes

There is no visible change, just a reduction in duplicate work.

## Checklist

- ~I have added appropriate error handling~ _N/A_
- ~I have added or updated documentation as needed~ _N/A_

## Additional context

I left behind some calls to `add_instrumentation_data` which could have been moved.

```ruby
def read_resource_no_content(request)
  add_instrumentation_data(resource_uri: request[:uri])
  []
end
```

I wasn't sure if we wanted _all_ reads to unconditionally instrument the URI.

```ruby
def call_tool(request)
  tool_name = request[:name]
  tool = tools[tool_name]
  unless tool
    add_instrumentation_data(error: :tool_not_found)
    raise RequestHandlerError.new("Tool not found #{tool_name}", request, error_type: :tool_not_found)
  end

  arguments = request[:arguments]
  add_instrumentation_data(tool_name:)

  # ...
end
```

I wasn't sure if we wanted to unconditionally instrument the `tool_name`, or only instrument _known_ tools (as we do now).

```ruby
def get_prompt(request)
  prompt_name = request[:name]
  prompt = @prompts[prompt_name]
  unless prompt
    add_instrumentation_data(error: :prompt_not_found)
    raise RequestHandlerError.new("Prompt not found #{prompt_name}", request, error_type: :prompt_not_found)
  end

  add_instrumentation_data(prompt_name:)

  # ...
end
```

Likewise, I wasn't sure if we only want to instrument _known_ prompts.

### DIFF
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -176,7 +176,6 @@ module MCP
     end
 
     def init(request)
-      add_instrumentation_data(method: Methods::INITIALIZE)
       {
         protocolVersion: configuration.protocol_version,
         capabilities: capabilities,
@@ -185,12 +184,10 @@ module MCP
     end
 
     def list_tools(request)
-      add_instrumentation_data(method: Methods::TOOLS_LIST)
       @tools.map { |_, tool| tool.to_h }
     end
 
     def call_tool(request)
-      add_instrumentation_data(method: Methods::TOOLS_CALL)
       tool_name = request[:name]
       tool = tools[tool_name]
       unless tool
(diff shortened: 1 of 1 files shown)


koic (approved):
I think it's reasonable to remove these calls to `add_instrumentation_data`, as they are redundant.

PR #67, opened 2025-06-19
Remove shebang from example files

koic:
## Motivation and Context

Since the example files are not executable binaries, this PR removes the shebangs.

This ensures that all examples are consistently executed via `ruby` command. It also removes the executable permission from `examples/stdio_server.rb` using `chmod 444`, bringing it in line with the other example files.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -9,7 +9,7 @@ A simple server that communicates over standard input/output. This is useful for
 
 **Usage:**
 ```bash
-ruby examples/stdio_server.rb
+$ ruby examples/stdio_server.rb
 {"jsonrpc":"2.0","id":0,"method":"tools/list"}
 ```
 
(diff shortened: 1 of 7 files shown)


examples/README.md
```diff
@@ -9,7 +9,7 @@ A simple server that communicates over standard input/output. This is useful for
 
 **Usage:**
 ```bash
-ruby examples/stdio_server.rb
+$ ruby examples/stdio_server.rb
```
sambostock, line 12: If we're using prefixes before running commands, we should use the `console` language for the codeblock

````markdown
```console
$ ruby examples/stdio_server.rb
````

(Same for the other code blocks)

koic: Yeah, I think so too. That said, this repository uses `bash`, so a separate PR will likely change it to `console`.

PR #68, opened 2025-06-19
Remove `json-schema` limit and reorganize dev deps

sambostock:
## Motivation and Context

We're currently limiting the `json-schema` gem to `~> 4.1`, which means applications using a `5.x` version get a dependency conflict. This addresses that.

While I was at it, I also moved all development dependencies to `Gemfile` for consistency.

## How Has This Been Tested?

No, but the only breaking change in `json-schema@5.0`'s changelog is dropping support for Ruby versions older than the minimum one we support.

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist

- ~I have added appropriate error handling~
- ~I have added or updated documentation as needed~

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,3 +4,6 @@ inherit_gem:
 plugins:
   - rubocop-minitest
   - rubocop-rake
+
+Gemspec/DevelopmentDependencies:
+  Enabled: true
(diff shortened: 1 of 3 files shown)


koic (approved):
I like this change. I was also considering moving the `add_development_dependency` entries to the Gemfile :-)

PR #73, opened 2025-07-11
fixes input schema validation for schemas without required fields

jkraemer:
Input Schemas that declare no required properties raised an ArgumentError due to the present but empty `required` field in `InputSchema#to_h`.

A test case reproducing the issue is included in the patch.


atesgoral (approved):

atesgoral:
You beat me to it. Thanks for fixing this!

PR #81, opened 2025-07-17
[tools] add support for structured responses

honzasterba:
Add new response class to be able to return structuredContent as part of tools response in a backwards compatible way

## Motivation and Context
Structured responses are an established part of the spec and the SDK should support it.

## How Has This Been Tested?
Add new tested and tested with internal MCP server implementation.

## Breaking Changes
None

## Types of changes
- [ ] Documentation update

## Checklist


dieter-medium:
Nice job! 

Does this not also require an 'outputSchema'?

honzasterba:
output schema is optional for structured output, for now I just added support for the response field

test/test_helper.rb
```diff
@@ -17,7 +17,7 @@
 
 require_relative "instrumentation_test_helper"
 
-Minitest::Reporters.use!(Minitest::Reporters::ProgressReporter.new)
+Minitest::Reporters.use!(Minitest::Reporters::ProgressReporter.new) unless ENV["RM_INFO"]
```
atesgoral: Can you explain where/when `RM_INFO` comes from? (Explicitly calling out this change in the PR description or punting it to its specific PR would be favourable.)

lib/mcp/tool/structured_response.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module MCP
+  class Tool
+    class StructuredResponse
+      attr_reader :content, :is_error
+
+      # @param structured_content [Hash] The structured content of the response, must be provided.
+      # @param content [String, nil] The content array of the response, can be nil. If nil will generate a single element with structured content converted to JSON string.
```
atesgoral, line 9: > For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

Not a MUST, but a SHOULD, but I agree with doing it for backwards compatibility in this gem.

test/test_helper.rb
honzasterba: created a new PR here https://github.com/modelcontextprotocol/ruby-sdk/pull/97/files

honzasterba:
separated unrelated change into new PR
improved commit message

honzasterba:
anything else needed to merge this?

koic:
According to [https://modelcontextprotocol.io/specification/2025-06-18/schema#calltoolresult](https://modelcontextprotocol.io/specification/2025-06-18/schema#calltoolresult), it seems preferable to add an optional `structured_content` keyword argument to the existing `MCP::Tool::Response` rather than creating a new `MCP::Tool::StructuredResponse`.

anandaroop:
Thank you for this PR @honzasterba.

I also require this functionality. 

Therefore I have tried another approach in line with @koic's suggestion at #147

koic:
I'll close this since I think the issue has been resolved in #147. Thank you for raising it!

PR #84, opened 2025-07-21
implements the notifications/initialized method

jkraemer:
According to https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle, "After successful initialization, the client MUST send an initialized notification to indicate it is ready to begin normal operations".

This implementation just replies with an empty response (same as ping).

## Motivation and Context

Claude Code assumed the MCP server wasn't working when this method was not handled gracefully.

## How Has This Been Tested?

test case added, works in my app.

## Types of changes

- [ ] Documentation update

### DIFF
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -74,6 +74,7 @@ module MCP
         Methods::PROMPTS_GET => method(:get_prompt),
         Methods::INITIALIZE => method(:init),
         Methods::PING => ->(_) { {} },
+        Methods::NOTIFICATIONS_INITIALIZED => ->(_) {},
 
         # No op handlers for currently unsupported methods
         Methods::RESOURCES_SUBSCRIBE => ->(_) {},
--- a/test/mcp/server_test.rb
+++ b/test/mcp/server_test.rb
@@ -160,6 +160,26 @@ module MCP
       assert_instrumentation_data({ method: "unsupported_method" })
     end
 
+    test "#handle notifications/initialized returns nil response" do
+      request = {
+        jsonrpc: "2.0",
+        method: "notifications/initialized",
+      }
+
+      assert_nil @server.handle(request)
+      assert_instrumentation_data({ method: "notifications/initialized" })
+    end
+
+    test "#handle_json notifications/initialized returns nil response" do
+      request = JSON.generate({
+        jsonrpc: "2.0",
+        method: "notifications/initialized",
+      })
+
+      assert_nil @server.handle_json(request)
+      assert_instrumentation_data({ method: "notifications/initialized" })
+    end
+
     test "#handle tools/list returns available tools" do
       request = {
         jsonrpc: "2.0",


koic (approved):
Yeah, the current behavior that results in the following error is problematic.

```
{jsonrpc: "2.0", id: "123", error: {code: -32601, message: "Method not found", data: "notifications/initialized"}}
```

It must be able to respond to `notifications/initialized` in accordance with the specification.

lib/mcp/server.rb
```diff
@@ -66,6 +66,7 @@ def initialize(
         Methods::PROMPTS_GET => method(:get_prompt),
         Methods::INITIALIZE => method(:init),
         Methods::PING => ->(_) { {} },
+        Methods::NOTIFICATIONS_INITIALIZED => ->(_) { {} },
```
atesgoral: Actually shouldn't this return `nil` (i.e. return nothing)? Notifications are special methods that don't have a response.

test/mcp/server_test.rb
```diff
@@ -141,6 +141,25 @@ class ServerTest < ActiveSupport::TestCase
       assert_instrumentation_data({ method: "unsupported_method" })
     end
 
+    test "#handle notifications/initialized returns empty response" do
+      request = {
+        jsonrpc: "2.0",
+        method: "notifications/initialized",
+        id: "123",
```
atesgoral: An `id` should not be present here, since this is a notification.

lib/mcp/server.rb
atesgoral: See https://www.jsonrpc.org/specification#notification

atesgoral (changes requested):
This should be implemented as an actual notification, not a method. See https://www.jsonrpc.org/specification#notification

jkraemer:
I have changed it to return nil and fixed the test case to not send an `id`. I did not find any code dedicated to handling incoming notifications outside the regular method call handling - do we leave it at that for now or would it make sense to have a separate handler for anything `notifications/*`?

atesgoral (approved):

atesgoral:
@jkraemer A single handler suffices because we delegate response handling to the underlying json_rpc_handler gem and it's responsible for handling nil results.

alexandru-calinoiu:
Also notice that the response code for this request should be 202 (:accepted). In order to have this working with claude.ai I had to implement the following unglines:

```ruby
    status, headers, body = @transport.handle_request(rack_request)

    if params[:method] == "notifications/initialized"
      status = 202
    end
```

K4sku:
It seems like Claude.ai is wrong here. 

Both rpc and MCP specs state that notification does not require a response. 
[However, python SDK does return 202](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/streamable_http.py#L375)

alexandru-calinoiu:
> It seems like Claude.ai is wrong here.
> 
> Both rpc and MCP specs state that notification does not require a response. [However, python SDK does return 202](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/streamable_http.py#L375)

I agree, but I am not sure how we can raise this with them.

K4sku:
There is a "give feedback" button in Claude Desktop settings; it opens a Google Form. 
I just submitted a bug there. I have passed a link to this PR.

jkraemer:
> It seems like Claude.ai is wrong here.
> 
> Both rpc and MCP specs state that notification does not require a response. [However, python SDK does return 202](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/streamable_http.py#L375)

I think we need to differentiate between JSON-RPC and HTTP here - even a 'no response' in JSON-RPC land requires *some* HTTP status code on the response if HTTP is the underlying protocol over which JSON-RPC messages are transported. I think a 202 with an empty body is entirely appropriate in this case.

K4sku:
> > It seems like Claude.ai is wrong here.
> > Both rpc and MCP specs state that notification does not require a response. [However, python SDK does return 202](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/server/streamable_http.py#L375)
> 
> I think we need to differentiate between JSON-RPC and HTTP here - even a 'no response' in JSON-RPC land requires _some_ HTTP status code on the response if HTTP is the underlying protocol over which JSON-RPC messages are transported. I think a 202 with an empty body is entirely appropriate in this case.

It probably should accept any "successful" HTTP status code, given that
1. It's not documented anywhere.
2. Claude Code and MCP Inspector are fine with 200.

PS. Sorry for the offtopic.

atesgoral:
I couldn't glean from claude.ai's HTML what framework it's built with (and didn't spend too much time with it), but both Claude Desktop and the Inspector are likely using the official TS SDK. We can check what the TS SDK client is doing.

I can also ask the steering committee what SDK claude.ai is using and why they chose to be strict about 202.

I agree that clients should just be happy about 20x.. And the Ruby SDK can return a 202 for all notifications, not just this one.

koic:
As for the response status, it is handled separately in #84 and #114, so I think this PR can be merged independently of them.

@jkraemer Can you squash your commits into one?

jkraemer:
> @jkraemer Can you squash your commits into one?

Done!

koic:
Thanks!

PR #85, opened 2025-07-22
Fix tools with no args

SheldonNunes:
## Motivation and Context

This PR fixes a bug in the ruby-sdk where if a tool provided has zero arguments it will throw an error. In order to resolve this the following line needed updating to handle `nil` case:
`args = arguments&.transform_keys(&:to_sym) || {}`

Example of tools not requiring args e.g. Listing a set of resources available / configuration.

## How Has This Been Tested?

Tested in local application and added unit test for this case

## Breaking Changes

No breaking change

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -314,7 +314,7 @@ module MCP
     end
 
     def call_tool_with_args(tool, arguments)
-      args = arguments.transform_keys(&:to_sym)
+      args = arguments&.transform_keys(&:to_sym) || {}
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -405,7 +405,7 @@ tool = MCP::Tool.define(
     title: "My Tool",
     read_only_hint: true
   }
-) do |args, server_context|
+) do
```
atesgoral: Let's not modify this single example of using `define()` with a block with args.

PR #87, opened 2025-07-23
Add `instructions` support to `MCP::Server`

koic:
## Motivation and Context

The response of the `init` method can include `instructions`. https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle

`instructions` appear to be supported starting from the 2025-03-26 specification.

- https://modelcontextprotocol.io/specification/2025-03-26/basic/lifecycle (supports `instructions`)
- https://modelcontextprotocol.io/specification/2024-11-05/basic/lifecycle (does not support `instructions`)

If `instructions` is used with an unsupported protocol version, an `ArgumentError` will be raised.

## How Has This Been Tested?

Existing tests have been updated and new tests have been added.

## Breaking Changes

None.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

As additional context, the TypeScript SDK also treats `instructions` as optional. https://github.com/modelcontextprotocol/typescript-sdk/blob/1.16.0/src/types.ts#L356-L361

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,11 +155,11 @@ You can use the `Server#handle_json` method to handle requests.
 
 ```ruby
 class ApplicationController < ActionController::Base
-
   def index
     server = MCP::Server.new(
       name: "my_server",
       version: "1.0.0",
+      instructions: "Use the tools of this server as a last resort",
       tools: [SomeTool, AnotherTool],
       prompts: [MyPrompt],
       server_context: { user_id: current_user.id },
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -31,11 +31,12 @@ module MCP
 
     include Instrumentation
 
-    attr_accessor :name, :version, :tools, :prompts, :resources, :server_context, :configuration, :capabilities, :transport
+    attr_accessor :name, :version, :instructions, :tools, :prompts, :resources, :server_context, :configuration, :capabilities, :transport
 
     def initialize(
       name: "model_context_protocol",
       version: DEFAULT_VERSION,
+      instructions: nil,
       tools: [],
       prompts: [],
       resources: [],
@@ -47,6 +48,7 @@ module MCP
     )
       @name = name
       @version = version
+      @instructions = instructions
       @tools = tools.to_h { |t| [t.name_value, t] }
       @prompts = prompts.to_h { |p| [p.name_value, p] }
       @resources = resources
@@ -54,6 +56,12 @@ module MCP
       @resource_index = index_resources_by_uri(resources)
       @server_context = server_context
       @configuration = MCP.configuration.merge(configuration)
+
+      if @configuration.protocol_version == "2024-11-05" && @instructions
+        message = "`instructions` supported by protocol version 2025-03-26 or higher"
+        raise ArgumentError, message
+      end
+
       @capabilities = capabilities || default_capabilities
 
       @handlers = {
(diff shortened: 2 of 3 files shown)


lib/mcp/server.rb
```diff
@@ -47,13 +48,19 @@ def initialize(
...
       @prompts = prompts.to_h { |p| [p.name_value, p] }
       @resources = resources
       @resource_templates = resource_templates
       @resource_index = index_resources_by_uri(resources)
       @server_context = server_context
       @configuration = MCP.configuration.merge(configuration)
+      if @configuration.protocol_version == "2024-11-05" && @instructions
+        message = "`instructions` supported by protocol version 2025-03-26 or higher"
+        raise ArgumentError, message
+      end
```
koic, line 63: If accepting the `instructions` parameter with the older "2024-11-05" protocol version is considered preferable, this logic could be removed as it only introduces unnecessary complexity.

README.md
```diff
@@ -155,11 +155,11 @@ You can use the `Server#handle_json` method to handle requests.
 
 ```ruby
 class ApplicationController < ActionController::Base
-
   def index
     server = MCP::Server.new(
       name: "my_server",
       version: "1.0.0",
+      instructions: "A simple MCP server with custom tools and prompts",
```
atesgoral: https://modelcontextprotocol.io/specification/2025-06-18/schema describes `instructions` as:

> Instructions describing how to use the server and its features.
>
> This can be used by clients to improve the LLM’s understanding of available tools, resources, etc. It can be thought of like a “hint” to the model. For example, this information MAY be added to the system prompt.

So, it's not meant as a "description" for the server that is intended for humans, but more of a "how to get the best out of me" hint intended for the LLM.

I'm having a hard time coming up with a good example.

Maybe something humourous like "Use this server wisely" or "Use the tools of this server as a last resort"... Just something sounding more like it's intended for the LLM.

lib/mcp/server.rb
atesgoral: In case there are picky clients that could error on unexpected properties, I think being safe (backwards compatible) is good here.

But are we doing this kind of check uniformly for all features? Is there a Ruby-friendly annotation technique we could adopt to start doing this more declaratively...

I'm happy with this change as it is, though.

lib/mcp/server.rb
```diff
@@ -47,13 +48,19 @@ def initialize(
...
       @version = version
+      @instructions = instructions
       @tools = tools.to_h { |t| [t.name_value, t] }
       @prompts = prompts.to_h { |p| [p.name_value, p] }
       @resources = resources
       @resource_templates = resource_templates
       @resource_index = index_resources_by_uri(resources)
       @server_context = server_context
       @configuration = MCP.configuration.merge(configuration)
+      if @configuration.protocol_version == "2024-11-05" && @instructions
```
atesgoral: Stylistic subjective nitpick:

```suggestion
      @configuration = MCP.configuration.merge(configuration)
      
      if @configuration.protocol_version == "2024-11-05" && @instructions
```

lib/mcp/server.rb
koic: At some point, it would be ideal to enable declarative descriptions using a DSL instead of procedural ones.

README.md
koic: "Use the tools of this server as a last resort" sounds like a cool example! I've updated it accordingly.

PR #93, opened 2025-07-26
Clarify resource read handler in README

andyw8:
This better matches the code in `/examples` and helps illustrate the usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -575,7 +575,7 @@ server.resources_read_handler do |params|
   [{
     uri: params[:uri],
     mimeType: "text/plain",
-    text: params[:uri],
+    text: "Hello from example resource! URI: #{params[:uri]}"
   }]
 end
 


README.md
```diff
@@ -575,7 +575,7 @@ server.resources_read_handler do |params|
   [{
     uri: params[:uri],
     mimeType: "text/plain",
-    text: params[:uri],
+    text: "Hello from example resource!"
```
koic: As noted in #47, based on my experience, returning a simple string like "Hello" from `resources/read` was confusing when I first used `resources/read`. It would be helpful to have a more illustrative example...

andyw8: I've updated it have both the message and the URI, which hopefully is clearer.

koic:
Let's proceed with this. Can you squash your commits into one?

koic (approved):

PR #94, opened 2025-07-26
Use valid syntax for URIs in examples and tests

andyw8:
The [spec](https://modelcontextprotocol.io/specification/2025-06-18/server/resources) appear to use only properly structured URIs, so the examples and tests should reflect that.


examples/http_server.rb
```diff
@@ -61,7 +61,7 @@ def template(args, server_context:)
   prompts: [ExamplePrompt],
   resources: [
     MCP::Resource.new(
-      uri: "test_resource",
+      uri: "example://test_resource",
```
koic: It might be preferable to use a more generic protocol name (e.g., `https://`). As for the example domain of the resource, something like `.invalid` (returns NXDOMAIN) could be considered.

- https://www.rfc-editor.org/rfc/rfc2606.html
- https://www.rfc-editor.org/rfc/rfc6761.html

For example, a value such as `https://test_resource.invalid` could be used.

test/mcp/server_test.rb
```diff
@@ -44,14 +44,14 @@ class ServerTest < ActiveSupport::TestCase
       end
 
       @resource = Resource.new(
-        uri: "test_resource",
+        uri: "test://test_resource",
```
koic: Ditto.

koic:
Can you also update the usage of `example://` and `test://` in the PR accordingly?

koic:
This update looks good. Can you squash your commits into one?

test/mcp/server/transports/stdio_notification_integration_test.rb
```diff
@@ -228,7 +228,7 @@ def puts(message)
           # Add a resource and notify
           @server.resources = [
             MCP::Resource.new(
-              uri: "test://resource",
+              uri: "example://test_resource",
```
andyw8, line 231: I had missed this one earlier.

atesgoral (approved):
TIL about the .invalid TLD. I've always been using example.com.

PR #97, opened 2025-07-28
[test] minites-reporters do not work with RubyMine

honzasterba:
Disable Minitest::Reporters when running tests in RubyMine

## Motivation and Context
I want to execute tests directly in ruby mine and Minitest::Reporters is not compatible with RubyMine test runner

## How Has This Been Tested?
Locally running tests in and outside of RubyMine

## Breaking Changes
none

## Types of changes
- [ ] Documentation update
- [X] Dev tools support

## Checklist
- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -3,7 +3,6 @@
 require "mcp"
 
 require "minitest/autorun"
-require "minitest/reporters"
 require "minitest/mock"
 require "mocha/minitest"
 
@@ -13,5 +12,3 @@ require "active_support/test_case"
 require "sorbet-runtime"
 
 require_relative "instrumentation_test_helper"
-
-Minitest::Reporters.use!(Minitest::Reporters::ProgressReporter.new)
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ gemspec
 
 # Specify development dependencies below
 gem "minitest", "~> 5.1", require: false
-gem "minitest-reporters"
 gem "mocha"
 
 gem "rubocop-minitest", require: false


test/test_helper.rb
```diff
@@ -17,7 +17,10 @@
 
 require_relative "instrumentation_test_helper"
 
-Minitest::Reporters.use!(Minitest::Reporters::ProgressReporter.new)
+unless ENV["RM_INFO"]
+  # RubyMine does not support Minitest reporters, so we use the default reporter
+  Minitest::Reporters.use!(Minitest::Reporters::ProgressReporter.new)
+end
```
koic: There is concern about the inclusion of the RubyMine-specific environment variable `"RM_INFO"`, which depends on the local environment. It may be worth reconsidering whether `Minitest::Reporters` configuration is necessary in the first place. If not, removing the use of `Minitest::Reporters` might be the simpler option.

honzasterba: `Minitest::Reporters` is definitely not critical to anything the test suite is doing, it's just an fancy progress bar when running the tests

koic: Yeah, if the appearance of the test progress isn't important, removing it might be the simplest solution.

honzasterba: per your suggestion removed the gem completely

koic (approved):
Looks good to me. However, since I'm not sure about the background of its introduction, I will leave the decision to @atesgoral or @topherbullock.

topherbullock:
> Looks good to me. However, since I'm not sure about the background of its introduction, I will leave the decision to @atesgoral or @topherbullock.

LGTM; this made sense when this was an internal gem running in Shopify's build system, but no longer!

Issue #99, opened 2025-07-29
Stateless MCP Servers


bigH:
**Is your feature request related to a problem? Please describe.**
The Ruby SDK currently stores sessions in memory, which prevents it from supporting stateless backends. This is a scalability issue because session state cannot be shared across nodes. In contrast, the Python SDK [supports a `stateless` configuration value](https://github.com/modelcontextprotocol/python-sdk/blob/959d4e39ae13e45d3059ec6d6ca82fb231039a91/src/mcp/server/streamable_http_manager.py#L52).

**Describe the solution you'd like**
We'd like the Ruby SDK to support stateless session management, similar to the Python SDK.

**Describe alternatives you've considered**
- using sticky sessions — this is not something we can pursue
- forking the library — not ideal, but something we can do to unblock
- switching to a different SDK — `fast-mcp` has no Streamable HTTP support, so this is a non-starter

**Additional context**
We use a minimum of 2 pods for all services to ensure availability. This feature is critical for us to use the Ruby SDK to build remote MCPs.

sbshah97:
Looks like this is part of the specification for the November release? Probably something to add during that time?

collimarco:
This gem is totally useless without this feature. No real Rails app can use this gem at the moment.

paulRbr:
Thanks for adding the `stateless` attribute to the SSE transport layer. However, I have a question on how do you use this lib to be able to do streaming responses and actually keep sessions in a production-like environment (I.e. more than 2 workers)?

Do you think the streamable http transport layer could be configurable to be able to receive any kind of “store” for the sessions (and associated streams) in place of the current memory hash? I'm not an expert in transport layers, so I'm not even sure we can do streamable http with distributed streams across web servers.

If you folks have any pointers on that, I'd be happy to help and contribute a PR to the lib 🙏 

Thanks for the help!

ghost:
Hi, this feature isn't working using the stock `mcp` gem from rubygems. `MCP::Server::Transports::StreamableHTTPTransport.new` only takes one argument so we can't pass in `stateless: true`.

koic:
The feature is included in the main branch, but not in the current release (v0.4.0). It will be available in the next release.

PR #101, opened 2025-07-31
Add `stateless` feature to Streamable HTTP protocol

bigH:
See: https://github.com/modelcontextprotocol/ruby-sdk/issues/99

Currently, this implementation of MCP requires that sessions are held in memory for Server-Sent Events (SSEs). The session data maps an ID to a live IO stream. This doesn't work for Chime for a few reasons:

- We run >1 replica, so in-memory storage of session state is not going to work
- When contacting a backend with no knowledge of a given session, the server responds with an error, breaking downstream agent functionality
- Chime's default configuration doesn't allow for long-lived connections and connections are terminated forcibly after a timeout

So, here we add tests and validate that `stateless` configuration behaves according to spec. In situations where the spec is unclear about something, we defer to the [Python SDK](https://github.com/modelcontextprotocol/python-sdk).

### DIFF
--- a/test/mcp/server/transports/streamable_http_transport_test.rb
+++ b/test/mcp/server/transports/streamable_http_transport_test.rb
@@ -573,6 +573,128 @@ module MCP
           assert_equal "Method not allowed", body["error"]
         end
 
+        test "stateless mode allows requests without session IDs, responding with no session ID" do
+          stateless_transport = StreamableHTTPTransport.new(@server, stateless: true)
+
+          init_request = create_rack_request(
+            "POST",
+            "/",
+            { "CONTENT_TYPE" => "application/json" },
+            { jsonrpc: "2.0", method: "initialize", id: "init" }.to_json,
+          )
+          init_response = stateless_transport.handle_request(init_request)
+          assert_nil init_response[1]["Mcp-Session-Id"]
+        end
+
+        test "stateless mode responds without any session ID when session ID is present" do
+          stateless_transport = StreamableHTTPTransport.new(@server, stateless: true)
+
+          request = create_rack_request(
+            "POST",
+            "/",
+            {
+              "CONTENT_TYPE" => "application/json",
+              "HTTP_MCP_SESSION_ID" => "unseen_session_id",
+            },
+            { jsonrpc: "2.0", method: "ping", id: "123" }.to_json,
+          )
+
+          response = stateless_transport.handle_request(request)
+          assert_equal 200, response[0]
+          assert_equal(
+            {
+              "Content-Type" => "application/json",
+            },
+            response[1],
+          )
+
+          body = JSON.parse(response[2][0])
+          assert_equal "2.0", body["jsonrpc"]
+          assert_equal "123", body["id"]
+        end
+
+        test "stateless mode responds with 405 when SSE is requested" do
+          stateless_transport = StreamableHTTPTransport.new(@server, stateless: true)
+
+          get_request = create_rack_request(
+            "GET",
+            "/",
+            {
+              "CONTENT_TYPE" => "application/json,text/event-stream",
+            },
+          )
+          response = stateless_transport.handle_request(get_request)
+          assert_equal 405, response[0]
+          assert_equal({ "Content-Type" => "application/json" }, response[1])
+
+          body = JSON.parse(response[2][0])
+          assert_equal "Method not allowed", body["error"]
+        end
+
+        test "stateless mode silently responds with success to session DELETE when session ID is not present" do
+          stateless_transport = StreamableHTTPTransport.new(@server, stateless: true)
+
+          delete_request = create_rack_request(
+            "DELETE",
+            "/",
+            {},
+          )
+          response = stateless_transport.handle_request(delete_request)
+          assert_equal 200, response[0]
+          assert_equal({ "Content-Type" => "application/json" }, response[1])
+
+          body = JSON.parse(response[2][0])
+          assert body["success"]
+        end
+
+        test "stateless mode silently responds with success to session DELETE when session ID is provided" do
+          stateless_transport = StreamableHTTPTransport.new(@server, stateless: true)
+
+          delete_request = create_rack_request(
+            "DELETE",
+            "/",
+            { "HTTP_MCP_SESSION_ID" => "session_id" },
+          )
+          response = stateless_transport.handle_request(delete_request)
+          assert_equal 200, response[0]
+          assert_equal({ "Content-Type" => "application/json" }, response[1])
+
+          body = JSON.parse(response[2][0])
+          assert body["success"]
+        end
+
+        test "stateless mode does not support server-sent events" do
+          stateless_transport = StreamableHTTPTransport.new(@server, stateless: true)
+
+          e = assert_raises(RuntimeError) do
+            stateless_transport.send_notification(
+              "test_notification",
+              { message: "Hello" },
+              session_id: "some_session_id",
+            )
+          end
+
+          assert_equal("Stateless mode does not support notifications", e.message)
+        end
+
+        test "stateless mode responds with 202 when client sends a notification/initialized request" do
+          stateless_transport = StreamableHTTPTransport.new(@server, stateless: true)
+
+          request = create_rack_request(
+            "POST",
+            "/",
+            { "CONTENT_TYPE" => "application/json" },
+            { jsonrpc: "2.0", method: "notifications/initialized" }.to_json,
+          )
+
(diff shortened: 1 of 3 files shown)


koic:
Does `handle_post` method need to handle the case for stateless?

bigH:
> Does `handle_post` method need to handle the case for stateless?

My understanding is that it does not. `handle_post` calls `handle_regular_request`, which [only uses SSE if a stream exists](https://github.com/modelcontextprotocol/ruby-sdk/blob/2c3f218d8f45d852d67b7327faff828e9e2df914/lib/mcp/server/transports/streamable_http_transport.rb#L195-L201).

I've tested this against my own service and it works properly. What would it take to get this merged?

koic:
Ah, I see. Thank you for the explanation. Can you squash your commits into one?

bigH:
Happy to.

bigH:
Rebased on to latest main and squashed my commits.

koic (approved):

bigH:
Bump @koic. I think this is good to review/merge.

test/mcp/server/transports/streamable_http_transport_test.rb
```diff
@@ -575,6 +575,126 @@ class StreamableHTTPTransportTest < ActiveSupport::TestCase
...
+        test "stateless mode does not support server-sent events" do
+          stateless_transport = StreamableHTTPTransport.new(@server, stateless: true)
+
+          assert_raises(RuntimeError, "Stateless mode does not support notifications") do
+            stateless_transport.send_notification(
+              "test_notification",
+              { message: "Hello" },
+              session_id: "some_session_id",
+            )
+          end
```
koic: If my understanding is correct, it would be as follows.

```suggestion
          e = assert_raises(RuntimeError) do
            stateless_transport.send_notification(
              "test_notification",
              { message: "Hello" },
              session_id: "some_session_id",
            )
          end

          assert_equal("Stateless mode does not support notifications", e.message)
```

test/mcp/server/transports/streamable_http_transport_test.rb
```diff
@@ -575,6 +575,126 @@ class StreamableHTTPTransportTest < ActiveSupport::TestCase
...
+            { "CONTENT_TYPE" => "application/json" },
+            { jsonrpc: "2.0", method: "notifications/initialized" }.to_json,
+          )
+
+          response = stateless_transport.handle_request(request)
+          assert_equal 202, response[0]
+          assert_equal({ "Content-Type" => "application/json" }, response[1])
+
+          body = response[2][0]
+          assert body.blank?
```
koic: That's just my two cents.

```suggestion
          assert_nil(body)
```

test/mcp/server/transports/streamable_http_transport_test.rb
```diff
@@ -575,6 +575,128 @@ class StreamableHTTPTransportTest < ActiveSupport::TestCase
...
+          request = create_rack_request(
+            "POST",
+            "/",
+            { "CONTENT_TYPE" => "application/json" },
+            { jsonrpc: "2.0", method: "notifications/initialized" }.to_json,
+          )
+
+          response = stateless_transport.handle_request(request)
+          assert_equal 202, response[0]
+          assert_equal({ "Content-Type" => "application/json" }, response[1])
```
koic: It seems that this assertion is failing due to the changes in https://github.com/modelcontextprotocol/ruby-sdk/pull/105. Can you rebase with the latest main branch and update this assertion as follows?

```suggestion
          assert_empty(response[1])
```

koic:
@bigH This looks good overall to me. Can you rebase on the latest main branch, make sure the tests pass, and squash the commits into a single one?

Ginja:
I tested this change on an internal MCP server and it looks good 👍

Ginja:
On further testing I may have found a bug with this feature, but only when specifically using [socketry/falcon](https://github.com/socketry/falcon) (vs Puma).

```
{
    "time": "2025-08-21T20:28:04+00:00",
    "severity": "error",
    "pid": 20,
    "oid": 1176,
    "fiber_id": 1192,
    "subject": "Protocol::Rack::Adapter::Rack31",
    "annotation": "Reading HTTP/1.1 requests for Async::HTTP::Protocol::HTTP1::Server.",
    "message": "undefined method 'bytesize' for nil",
    "event": {
        "type": "failure",
        "root": "/app",
        "class": "NoMethodError",
        "message": "undefined method 'bytesize' for nil",
        "backtrace": [
            "/artifacts/bundle/ruby/3.4.0/gems/protocol-rack-0.16.0/lib/protocol/rack/body/enumerable.rb:27:in 'Array#sum'",
            "/artifacts/bundle/ruby/3.4.0/gems/protocol-rack-0.16.0/lib/protocol/rack/body/enumerable.rb:27:in 'Protocol::Rack::Body::Enumerable.wrap'",
            "/artifacts/bundle/ruby/3.4.0/gems/protocol-rack-0.16.0/lib/protocol/rack/body.rb:64:in 'Protocol::Rack::Body.wrap'",
            "/artifacts/bundle/ruby/3.4.0/gems/protocol-rack-0.16.0/lib/protocol/rack/response.rb:53:in 'Protocol::Rack::Response.wrap'",
            "/artifacts/bundle/ruby/3.4.0/gems/protocol-rack-0.16.0/lib/protocol/rack/adapter/generic.rb:160:in 'Protocol::Rack::Adapter::Generic#call'",
            "/artifacts/bundle/ruby/3.4.0/gems/protocol-http-0.52.0/lib/protocol/http/middleware.rb:53:in 'Protocol::HTTP::Middleware#call'",
            "/artifacts/bundle/ruby/3.4.0/gems/protocol-http-0.52.0/lib/protocol/http/content_encoding.rb:40:in 'Protocol::HTTP::ContentEncoding#call'",
            "/artifacts/bundle/ruby/3.4.0/gems/protocol-http-0.52.0/lib/protocol/http/middleware.rb:53:in 'Protocol::HTTP::Middleware#call'",
            "/artifacts/bundle/ruby/3.4.0/gems/falcon-0.52.3/lib/falcon/server.rb:66:in 'Falcon::Server#call'",
            "/artifacts/bundle/ruby/3.4.0/gems/async-http-0.91.0/lib/async/http/server.rb:57:in 'block in Async::HTTP::Server#accept'",
            "/artifacts/bundle/ruby/3.4.0/gems/async-http-0.91.0/lib/async/http/protocol/http1/server.rb:72:in 'Async::HTTP::Protocol::HTTP1::Server#each'",
            "/artifacts/bundle/ruby/3.4.0/gems/async-http-0.91.0/lib/async/http/server.rb:49:in 'Async::HTTP::Server#accept'",
            "/artifacts/bundle/ruby/3.4.0/gems/falcon-0.52.3/lib/falcon/server.rb:57:in 'Falcon::Server#accept'",
            "/artifacts/bundle/ruby/3.4.0/gems/io-endpoint-0.15.2/lib/io/endpoint/wrapper.rb:216:in 'block (2 levels) in IO::Endpoint::Wrapper#accept'",
            "/artifacts/bundle/ruby/3.4.0/gems/async-2.27.4/lib/async/task.rb:183:in 'block in Async::Task#run'",
            "/artifacts/bundle/ruby/3.4.0/gems/async-2.27.4/lib/async/task.rb:427:in 'block in Async::Task#schedule'"
        ]
    }
}
```

Digging into it, it looks like `nil` is not a valid response with Falcon. Adding the following to `streamable_http_transport.rb` fixes the issue. Something we may want to consider for those using Falcon.

```ruby
        def handle_regular_request(body_string, session_id)
          unless @stateless
            # If session ID is provided, but not in the sessions hash, return an error
            if session_id && !@sessions.key?(session_id)
              return [400, { "Content-Type" => "application/json" }, [{ error: "Invalid session ID" }.to_json]]
            end
          end

          # Ensure we always have a valid response body - convert nil to empty string
          response = @server.handle_json(body_string) || ""
```

bigH:
I'll come back to this next week and handle what you shared @Ginja @koic

koic:
Can you add user-facing documentation to the README?

Ginja:
@bigH any chance you have time to revisit this?

koic:
@bigH ping.

erickreutz:
We would love to see this merged!

bigH:
I'm pretty busy, but will take a look at finalizing this during the weekend.

koic:
@bigH Can you add documentation for this feature to the README.md and finally squash commits into one?

bigH:
@Ginja @koic applied both your suggestions, added docs and rebased. Everything seems to be passing.

PR #103, opened 2025-08-04
Add `logging` support

dak2:
## Motivation and Context

A server can send structured logging messages to the client. https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#logging

Logging was specified in the 2024-11-05 specification, but since it was not supported in ruby-sdk, I implemented it. https://modelcontextprotocol.io/specification/2024-11-05/server/utilities/logging

I also made it possible to output a simple notification message in the examples.

## How Has This Been Tested?

- [x] Check existing and added test cases for this case
- [x] Check a sse server of examples can send logging message to the client

## Breaking Changes

None

## Types of changes

- [x] Documentation update

## Checklist


lib/mcp/logging_message_notification.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require "json_rpc_handler"
+
+module MCP
+  class LoggingMessageNotification
+    VALID_LEVELS = ["debug", "info", "notice", "warning", "error", "critical", "alert", "emergency"].freeze
```
koic: Can you rename to the folliwng?

```suggestion
    LOG_LEVELS = ["debug", "info", "notice", "warning", "error", "critical", "alert", "emergency"].freeze
```

lib/mcp/server.rb
```diff
@@ -55,6 +56,7 @@ def initialize(
       @server_context = server_context
       @configuration = MCP.configuration.merge(configuration)
       @capabilities = capabilities || default_capabilities
+      @logging_message_notification = nil
```
koic, line 82: If I understand correctly, the Python SDK uses "info" level as the default. What do you think about doing the same?
https://github.com/modelcontextprotocol/python-sdk/blob/v1.12.3/src/mcp/server/fastmcp/server.py#L132

dak2: I don't think it's necessary to set a default value, but what do you think?

The log_level literal specified in the MCP spec appears to be defined in mcp/types.py, and it seems that no default value has been set.

https://github.com/modelcontextprotocol/python-sdk/blob/68e25d478b3b6a026b2d9a30b3e5f34f3b1290de/src/mcp/types.py#L905

The log_level in fastmcp/server.py#L132 appears to set the default value for uvicorn's log_level.

However, if this literal is the same as the one specified in the MCP spec, I don't think it meets the logging specifications, as levels such as `emergency` and `notice` are not defined.

lib/mcp/logging_message_notification.rb
dak2: Thanks! I fixed it.

https://github.com/modelcontextprotocol/ruby-sdk/compare/fd92524ef97afed38ba160e3d65506a7b9234f3c..6f91a19d0af8ac4cdd4e3cb0d7bcba1788e67d42

lib/mcp/server.rb
koic: That's true. There's no need to set something that's not explicitly specified in the specification.
Your current suggestion makes sense to me.

test/mcp/logging_message_notification_test.rb
```diff
@@ -0,0 +1,48 @@
+# typed: true
```
koic: Wouldn't the test pass even without this `# typed: true`?

dak2: Oh, it's probably a copy-paste mistake. Deleted.

https://github.com/modelcontextprotocol/ruby-sdk/compare/754f8cdc71113736b517f323b24f1987ea04c854..aa15a5bff45300b6dce18a97d7c6e74cbcd83837

koic:
@dak2 Oops, can you rebase with the latest master to resolve the conflicts?

dak2:
@koic 
Thank you for letting me know!
The conflict has been resolved. Could you please review it again?

koic (dismissed):

lib/mcp/logging_message_notification.rb
pusewicz: I'd suggest to also use the string array directly for easier parsing:

```ruby
LOG_LEVELS = %w[debug info notice warning error critical alert emergency].freeze
```

dak2: You are right according to [the Ruby style guide](https://rubystyle.guide/#percent-w).
However, I don't think it needs to be fixed.

This repository depends on [rubocop-shopify](https://github.com/Shopify/ruby-style-guide). It disables the `Style/WordArray` cop.
See: https://github.com/Shopify/ruby-style-guide/blob/0de5b278c576ad2f537f9f51f5897587a9b33841/rubocop.yml#L1405-L1407

Additionally, there don't seem to be any compelling reasons to enable it for this case in this repository.

koic:
Oops, sorry if my understanding of the specification and this implementation is different. As I understand it, with `logging/setLevel`, only messages at the specified level or higher should be sent from the server to the client as `notifications/message`. In other words, if the level `error` is specified, messages at `info` or `warning` level should not be notified.
I believe "Only sends error level and above" in the "Message Flow" section corresponds to this.
[https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#message-flow](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#message-flow)

It seems to me that the current PR is not implemented in this way. Could you check?

dak2:
@koic 
Thank you for checking it!

Oops, I missed it! That's true. This implementation doesn't meet a logging specification.
 I'll implement it soon when I have free time.

dak2:
@koic 

I implemented the logic to determine whether or not to send a notification based on level.
Could you please review it again?

https://github.com/modelcontextprotocol/ruby-sdk/compare/ef5049bab808dec719765a771c04e7c897c2a8a3..c9227469f523c930ab98b49567abf4dbd7213abe

If the notification level does not match, it returns nil. 
According to [the specification](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#message-flow), there are no restrictions on notifications when the level does not match, and the [java-sdk](https://github.com/modelcontextprotocol/java-sdk/blob/713ee1add0e29d184224aabdf06d024ef30a2754/mcp/src/main/java/io/modelcontextprotocol/server/McpAsyncServer.java#L773-L775) and [go-sdk](https://github.com/modelcontextprotocol/go-sdk/blob/1afdb1f27309c888be8a4bbb91deb584ede164be/mcp/server.go#L717-L719) implementations appear to be the same.

lib/mcp/logging_message_notification.rb
```diff
@@ -0,0 +1,46 @@
...
+    LOG_LEVELS = {
+      "emergency" => 0,
+      "alert" => 1,
+      "critical" => 2,
+      "error" => 3,
+      "warning" => 4,
+      "notice" => 5,
+      "info" => 6,
+      "debug" => 7,
+    }.freeze
```
koic, line 16: Intuitively I think `debug` with the lowest level of information should be ordered as `0` and `emergency` with the highest level of information as `7`. In other words, the keys are in reverse order.

lib/mcp/server.rb
```diff
@@ -138,6 +140,22 @@ def notify_resources_list_changed
       report_exception(e, { notification: "resources_list_changed" })
     end
 
+    def notify_logging_message(notification_level: nil, logger: nil, data: nil)
```
koic: `notify_log_message` looks a bit simple. Was there any SDK or other reference you based the naming on? Also, `level` keyword might be simpler for users than the `notification_level` keyword argument.


And, it looks like everything except the `logger` keyword is required. As for the ordering, the schema's `data`, `level`, `logger` sequence seems easier to understand.
https://modelcontextprotocol.io/specification/2025-06-18/schema#notifications%2Fmessage

test/mcp/logging_message_notification_test.rb
```diff
@@ -0,0 +1,63 @@
...
+      assert logging_message_notification.should_notify?(notification_level: "warning")
+      assert logging_message_notification.should_notify?(notification_level: "error")
+      assert logging_message_notification.should_notify?(notification_level: "critical")
+      assert logging_message_notification.should_notify?(notification_level: "alert")
+      assert logging_message_notification.should_notify?(notification_level: "emergency")
+    end
+
+    test "should_notify? returns false when notification level is lower priority than threshold level" do
+      logging_message_notification = LoggingMessageNotification.new(level: "warning")
+      assert_not logging_message_notification.should_notify?(notification_level: "notice")
```
koic: That's just my two cents. `refute` appears to be preferred to `assert_not` in this repository.

lib/mcp/server.rb
```diff
@@ -138,6 +140,22 @@ def notify_resources_list_changed
...
+      return unless @transport
+      raise LoggingMessageNotification::NotSpecifiedLevelError unless logging_message_notification&.level
+
+      current_level = notification_level || logging_message_notification.level
```
koic: `current_level` would be more precise as `log_level`. And  `log_level` is a required parameter, I didn't understand the intention of falling back to the global `logging_message_notification.level`.

lib/mcp/server.rb
```diff
@@ -138,6 +140,22 @@ def notify_resources_list_changed
...
+      return unless logging_message_notification.should_notify?(notification_level: current_level)
+
+      params = { level: current_level }
+      params[:logger] = logger if logger
```
koic: It seems there are no test cases for the `logger` keyword. Can you add it?

lib/mcp/server.rb
```diff
@@ -138,6 +140,22 @@ def notify_resources_list_changed
       report_exception(e, { notification: "resources_list_changed" })
     end
 
+    def notify_logging_message(notification_level: nil, logger: nil, data: nil)
+      return unless @transport
+      raise LoggingMessageNotification::NotSpecifiedLevelError unless logging_message_notification&.level
+
+      current_level = notification_level || logging_message_notification.level
+      return unless logging_message_notification.should_notify?(notification_level: current_level)
```
koic: `notification` is already clear from the receiver name, so `level` alone might be sufficient.


```suggestion
      return unless logging_message_notification.should_notify?(level: log_level)
```

koic:
User-facing documentation should be included in the README.

lib/mcp/logging_message_notification.rb
dak2: I had the same thought, but I prioritized meeting the specifications of RFC 5424's Numerical Code explicitly. 
However, looking at the [java-sdk](https://github.com/modelcontextprotocol/java-sdk/blob/cc67d8fa761d9122e03cc3c06d427739a6c0c633/mcp/src/main/java/io/modelcontextprotocol/spec/McpSchema.java#L2280-L2290), it seems to be defined in reverse order. 
Since that order is intuitive and easy to understand, I made the same correction.

lib/mcp/server.rb
dak2: > notify_log_message looks a bit simple. Was there any SDK or other reference you based the naming on?

That may be true. I also felt it was a little redundant when I implemented it. I referred to [typescript-sdk](https://github.com/modelcontextprotocol/typescript-sdk/blob/3bc2235d747c320dfa0b6227cc84414c6d0add89/src/server/index.ts#L362-L364) implementation. (Although the prefix is `send`)
However, as you said, notify_log_message is simpler, so I changed it.

It looks like [python-sdk](https://github.com/modelcontextprotocol/python-sdk/blob/eaf7cf41d5c1d3bcfd4416494ab5dc62b9c17a4a/src/mcp/server/session.py#L178-L198) uses the name `send_log_message`.

dak2: > Also, level keyword might be simpler for users than the notification_level keyword argument.
> And, it looks like everything except the logger keyword is required. As for the ordering, the schema's data, level, logger sequence seems easier to understand.

Fixed them.

test/mcp/logging_message_notification_test.rb
dak2: That seems to be the case. I don't have any particular preferences, so I'll just go with the style of the project.
Fixed it.

lib/mcp/server.rb
dak2: As you said, I couldn't test it. Added.

lib/mcp/server.rb
dak2: If it is defined as `level` name, I think it will be difficult to distinguish the KW argument name of `should_notify?` from the level instance variable name in `lib/mcp/logging_message_notification.rb`.

As an alternative, how about setting the KW argument name of `should_notify?` to `log_level`?
I think `log_level` is easier to understand because it indicates data received when notifying the log. What do you think?

lib/mcp/server.rb
dak2: The KW argument for level was implemented with a default value of nil, so it seems that it was implemented to falling back. I removed it because it was unnecessary.

dak2:
@koic 

> User-facing documentation should be included in the README.

I wrote how to use notifications/message in the README.md, but I would like feedback on whether the text is easy to understand.

lib/mcp/logging_message_notification.rb
koic: It seems my explanation was not clear. What is actually expected is the following.

```ruby
    LOG_LEVELS = {
      "debug" => 0,
      "info" => 1,
      "notice" => 2,
      "warning" => 3,
      "error" => 4,
      "critical" => 5,
      "alert" => 6,
      "emergency" => 7,
    }.freeze
```

README.md
```diff
@@ -119,6 +120,28 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
+- `notifications/message`
+
+#### Notification Logging Message Flow
+
+The `notifications/message` notification is used for structured logging between client and server.
+
+1. **Client sends logging configuration**: The client first sends a `logging/setLevel` request to configure the desired log level.
+2. **Server processes and notifies**: Upon receiving the log level configuration, the server uses `notify_log_message` to send log messages at the configured level and higher priority levels.For example, if "error" is configured, the server can send "error", "critical", "alert", and "emergency" messages. Please refer to `lib/mcp/logging_message_notification.rb` for log priorities in details.
+
+##### Usage Example
```
koic:
```suggestion
**Usage Example:**
```

README.md
```diff
@@ -119,6 +120,28 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
+
+```ruby
+# Client sets logging level
+# Request: { "jsonrpc": "2.0", "method": "logging/setLevel", "params": { "level": "error" } }
+
+# Server sends notifications for log events
+server.notify_log_message(
+  level: "error",
+  data: { error: "Connection Failed" },
+  logger: "DatabaseLogger"
```
koic:
```suggestion
  data: { error: "Connection Failed" },
  level: "error",
  logger: "DatabaseLogger"
```

lib/mcp/logging_message_notification.rb
```diff
@@ -0,0 +1,46 @@
...
+      "info" => 1,
+      "debug" => 0,
+    }.freeze
+
+    attr_reader :level
+
+    class InvalidLevelError < StandardError
+      def initialize
+        super("Invalid log level provided. Valid levels are: #{LOG_LEVELS.keys.join(", ")}")
+        @code = JsonRpcHandler::ErrorCode::InvalidParams
```
koic: What use cases is this `@code` set for the user?

lib/mcp/logging_message_notification.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+    end
+
+    class NotSpecifiedLevelError < StandardError
+      def initialize
+        super("Log level not specified. Please set a valid log level.")
+        @code = JsonRpcHandler::ErrorCode::InternalError
```
koic: Ditto.

README.md
```diff
@@ -119,6 +120,28 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
+The `notifications/message` notification is used for structured logging between client and server.
+
+1. **Client sends logging configuration**: The client first sends a `logging/setLevel` request to configure the desired log level.
+2. **Server processes and notifies**: Upon receiving the log level configuration, the server uses `notify_log_message` to send log messages at the configured level and higher priority levels.For example, if "error" is configured, the server can send "error", "critical", "alert", and "emergency" messages. Please refer to `lib/mcp/logging_message_notification.rb` for log priorities in details.
+
+##### Usage Example
+
+```ruby
+# Client sets logging level
+# Request: { "jsonrpc": "2.0", "method": "logging/setLevel", "params": { "level": "error" } }
```
koic: It would be better to show how to set the logging level in the Ruby SDK with Ruby code.

README.md
```diff
@@ -119,6 +120,28 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
 - `notifications/prompts/list_changed`
 - `notifications/resources/list_changed`
+- `notifications/message`
+
+#### Notification Logging Message Flow
+
+The `notifications/message` notification is used for structured logging between client and server.
+
+1. **Client sends logging configuration**: The client first sends a `logging/setLevel` request to configure the desired log level.
+2. **Server processes and notifies**: Upon receiving the log level configuration, the server uses `notify_log_message` to send log messages at the configured level and higher priority levels.For example, if "error" is configured, the server can send "error", "critical", "alert", and "emergency" messages. Please refer to `lib/mcp/logging_message_notification.rb` for log priorities in details.
```
koic: It would be good to provide an explanation of the Ruby API as part of the Ruby SDK. As for the MCP specification, I think it would be helpful to include a link to https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging.

lib/mcp/server.rb
koic: Ah, it doesn't really seem necessary for it to be a keyword argument in the first place. How about making it a positional argument instead?

dak2: Thank you. You're absolutely right. Fixed.

lib/mcp/logging_message_notification.rb
dak2: That's more intuitive. Fixed.

lib/mcp/logging_message_notification.rb
dak2: Its purpose is to communicate the log level is invalid. However, I realized that it must be returned as a JSON-RPC error object. This needs to be fixed.

https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#error-handling
https://www.jsonrpc.org/specification#error_object

lib/mcp/logging_message_notification.rb
dak2: Its purpose is to communicate the configuration error. This is also the same issue.
https://github.com/modelcontextprotocol/ruby-sdk/pull/103#discussion_r2302483116

lib/mcp/logging_message_notification.rb
dak2: Upon further review, it appears that the JSON-RPC error objects are returned and handled in https://github.com/Shopify/json-rpc-handler. 

I'm considering writing a patch for `json-rpc-handler` that explicitly calls the error objects `-32602` for invalid log levels and `-32603` for configuration errors.

The As it stands now: https://github.com/modelcontextprotocol/ruby-sdk/pull/103/files#diff-140fb0b13518d178300d879c19c9751cdd1a69f9c4ce07d3fb317b2c801b7d80R1109 expects `-32602`, but actually returns `-32603`.

It appears that this is likely being rescued by https://github.com/Shopify/json-rpc-handler/blob/a8a632681e24e53223a669612987686662975039/lib/json_rpc_handler.rb#L105-L111, resulting in `-32603` being returned.

When an invalid log level is encountered, [the MCP specification](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#error-handling) requires returning an error object with code `-32602`. However, it currently appears there is no way to specify and return `-32602` within json-rpc-handler. (I can write it to return the error object directly in the `ruby-sdk`, but delegating it to `json-rpc-handler` makes the responsibilities appear simpler.)

dak2: Hi, @atesgoral.

I submitted a [PR](https://github.com/Shopify/json-rpc-handler/pull/18) to the `json-rpc-handler` gem to enable it to handle JSON-RPC error objects. It seems to fulfill [the logging specification error handling requirements](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#error-handling). Could you please review it?

I created it because I believe handling JSON-RPC error objects is the responsibility of the `json-rpc-handler` gem.

koic:
@dak2 FYI, the [json_rpc_handler](https://github.com/Shopify/json-rpc-handler) was merged into this ruby-sdk repository in #175.

lib/mcp/logging_message_notification.rb
dak2: @koic 

Error handling for logging is now implemented in `lib/json_rpc_handler.rb`.
https://github.com/modelcontextprotocol/ruby-sdk/pull/103/files#diff-634a81d1c51c73876226353c172d04a666501f2c734c8bb923539e2e67faaa0dR108-R136
https://github.com/modelcontextprotocol/ruby-sdk/pull/103/files#diff-c43139a3840dc6d9991625febd7d3a950622b99ff41619bd6f91ca8d8c3e0954R270-R281

Some time has passed since the last update, so I'll reiterate the reason for the change. 
Servers must return a `JSON-RPC` error object when the log level or configuration is invalid. https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#error-handling
So, I implemented a feature that handles the `JSON-RPC` error object. (Maybe it would be better to split the PR.)

Referenced, [modelcontextprotocol/typescript-sdk](https://github.com/modelcontextprotocol/typescript-sdk/blob/783d53be1049041cc1f0a23c070c3400208a94ff/src/client/index.ts#L222-L224) and [modelcontextprotocol/python-sdk](https://github.com/modelcontextprotocol/python-sdk/blob/202af49857e452cdb8b55aa23310df8154e5b292/src/mcp/server/lowlevel/server.py#L374-L376).

README.md
dak2: Fixed.
https://github.com/modelcontextprotocol/ruby-sdk/pull/103/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R158-R173

README.md
dak2: Fixed.
https://github.com/modelcontextprotocol/ruby-sdk/pull/103/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R186-R193

README.md
dak2: I wrote a client-side implementation example using Ruby code. What do you think of this?
https://github.com/modelcontextprotocol/ruby-sdk/pull/103/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R165-R173

lib/mcp/logging_message_notification.rb
dak2: @atesgoral 

Merging json_rpc_handler into modelcontextprotocol/ruby-sdk is a huge help!!! Thanks, ates!
The above PR has been closed.

README.md
dak2: I rewrote README.md in details. How about this?
https://github.com/modelcontextprotocol/ruby-sdk/pull/103/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

dak2:
@koic 

Thanks for pointing that out!!
I've made a few commits.

lib/json_rpc_handler.rb
```diff
@@ -105,6 +105,8 @@ def process_request(request, id_validation_pattern:, &method_finder)
       result = method.call(params)
 
       success_response(id:, result:)
+    rescue MCP::Server::RequestHandlerError => e
```
dak2, line 108: I'm slightly concerned about the circular reference between `lib/json_rpc_handler.rb` and `lib/mcp/server.rb`, but I consider it to be fundamentally tied to the mechanism that handles server requests. Therefore, I think it's acceptable to be aware of RequestHandlerError.

lib/mcp/logging_message_notification.rb
```diff
@@ -0,0 +1,32 @@
...
+      @level = level
+    end
+
+    def valid_level?
+      LOG_LEVEL_SEVERITY.keys.include?(level)
+    end
+
+    def should_notify?(log_level)
+      LOG_LEVEL_SEVERITY[log_level] >= LOG_LEVEL_SEVERITY[level]
+    end
```
koic: Can you use the instance variable instead of private `attr_reader` method.

```suggestion
    def initialize(level:)
      @level = level
    end

    def valid_level?
      LOG_LEVEL_SEVERITY.keys.include?(@level)
    end

    def should_notify?(log_level)
      LOG_LEVEL_SEVERITY[log_level] >= LOG_LEVEL_SEVERITY[@level]
    end
```

lib/mcp/logging_message_notification.rb
```diff
@@ -0,0 +1,32 @@
...
+    def initialize(level:)
+      @level = level
+    end
+
+    def valid_level?
+      LOG_LEVEL_SEVERITY.keys.include?(level)
+    end
+
+    def should_notify?(log_level)
+      LOG_LEVEL_SEVERITY[log_level] >= LOG_LEVEL_SEVERITY[level]
```
koic: This would raise an error when `log_level` is nil, right? If so, please add a test and guard against it.

README.md
```diff
@@ -121,6 +122,83 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
+    error: "Database connection failed",
+    details: { host: "localhost", port: 5432 }
+  },
+  level: "error",
+  logger: "DatabaseLogger"  # Optional logger name
+)
+```
+
+**Key Features:**
+- Supports 8 log levels (debug, info, notice, warning, error, critical, alert, emergency) based on https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#log-levels
```
koic, line 196:
```suggestion
**Key Features:**

- Supports 8 log levels (debug, info, notice, warning, error, critical, alert, emergency) based on https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#log-levels
```

dak2: Addressed, thanks!
https://github.com/modelcontextprotocol/ruby-sdk/compare/17d7c25338b42f406b0f4e9288beb0103b44efeb..cdeb72c534f7c22bc576c299ae37b3b1ffee980b

lib/mcp/logging_message_notification.rb
dak2: Applied it, Thanks!
https://github.com/modelcontextprotocol/ruby-sdk/compare/272d442819f278d5b7e03cbb700c497da94f208d..17d7c25338b42f406b0f4e9288beb0103b44efeb

lib/mcp/logging_message_notification.rb
dak2: I guess it's fine to return false instead of throwing an error. 

Since [this method is called during the `notifications/message` phase](https://github.com/modelcontextprotocol/ruby-sdk/pull/103/changes#diff-c43139a3840dc6d9991625febd7d3a950622b99ff41619bd6f91ca8d8c3e0954R167-R169), returning false means that there is no need to respond to the client. According to [`JSON-RPC` 2.0 spec](https://www.jsonrpc.org/specification#notification), Notifications must not respond to the client.

Fixed.
https://github.com/modelcontextprotocol/ruby-sdk/pull/103/changes#diff-c43139a3840dc6d9991625febd7d3a950622b99ff41619bd6f91ca8d8c3e0954R169

README.md
```diff
@@ -121,6 +122,84 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
+| Level | Severity | Description |
+|-------|----------|-------------|
+| `debug` | 0 | Detailed debugging information |
+| `info` | 1 | General informational messages |
+| `notice` | 2 | Normal but significant events |
+| `warning` | 3 | Warning conditions |
+| `error` | 4 | Error conditions |
+| `critical` | 5 | Critical conditions |
+| `alert` | 6 | Action must be taken immediately |
+| `emergency` | 7 | System is unusable |
```
koic: The `Severity` values (i.e., from `0` to `7`) appear to be an internal implementation detail and not something users need to be aware of. How about renaming the `Level` heading to `Severity Level` and removing the `Severity` column?

README.md
```diff
@@ -121,6 +122,84 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
+
+#### How Logging Works
+
+1. **Client Configuration**: The client sends a `logging/setLevel` request to configure the minimum log level
+2. **Server Filtering**: The server only sends log messages at the configured level or higher severity
+3. **Notification Delivery**: Log messages are sent as `notifications/message` to the client
+
+For example, if the client sets the level to `"error"` (severity 4), the server will send messages with levels: `error`, `critical`, `alert`, and `emergency`.
+
+For more details, see the [MCP Logging specification](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging).
```
koic: FYI: https://github.com/modelcontextprotocol/ruby-sdk/pull/189

Can you update this URL to the following?

```suggestion
For more details, see the [MCP Logging specification](https://modelcontextprotocol.io/specification/latest/server/utilities/logging).
```

README.md
```diff
@@ -121,6 +122,84 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
+    details: { host: "localhost", port: 5432 }
+  },
+  level: "error",
+  logger: "DatabaseLogger"  # Optional logger name
+)
+```
+
+**Key Features:**
+
+- Supports 8 log levels (debug, info, notice, warning, error, critical, alert, emergency) based on https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#log-levels
```
koic, line 196: Ditto.

```suggestion
- Supports 8 log levels (debug, info, notice, warning, error, critical, alert, emergency) based on https://modelcontextprotocol.io/specification/latest/server/utilities/logging#log-levels
```

README.md
```diff
@@ -121,6 +122,84 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
+  level: "warning"
+)
+
+server.notify_log_message(
+  data: {
+    error: "Database connection failed",
+    details: { host: "localhost", port: 5432 }
+  },
+  level: "error",
+  logger: "DatabaseLogger"  # Optional logger name
```
koic: Nits.

```suggestion
  logger: "DatabaseLogger" # Optional logger name
```

README.md
dak2: I think so, addressed. Thanks!
https://github.com/modelcontextprotocol/ruby-sdk/compare/bdb005613864fc0ed2664b4e9fc0fb16e7bf8f2d..ac354523e320a9e5f39948bca653152437266119

README.md
dak2: Oops, I wasn't paying attention. I fixed it.
https://github.com/modelcontextprotocol/ruby-sdk/compare/bdb005613864fc0ed2664b4e9fc0fb16e7bf8f2d..ac354523e320a9e5f39948bca653152437266119

README.md
dak2: Fixed.
https://github.com/modelcontextprotocol/ruby-sdk/compare/bdb005613864fc0ed2664b4e9fc0fb16e7bf8f2d..ac354523e320a9e5f39948bca653152437266119

README.md
dak2: I've fixed that too.
https://github.com/modelcontextprotocol/ruby-sdk/compare/bdb005613864fc0ed2664b4e9fc0fb16e7bf8f2d..ac354523e320a9e5f39948bca653152437266119

README.md
```diff
@@ -121,6 +122,84 @@ Notifications follow the JSON-RPC 2.0 specification and use these method names:
...
+| Severity Level | Description |
+|-------|-------------|
+| `debug` | Detailed debugging information |
+| `info` | General informational messages |
+| `notice` | Normal but significant events |
+| `warning` | Warning conditions |
+| `error` | Error conditions |
+| `critical` | Critical conditions |
+| `alert` | Action must be taken immediately |
+| `emergency` | System is unusable |
```
koic: Thanks for taking care of this. Since there are now only two items, a list seems sufficient. This also looks better from an accessibility perspective.

```suggestion
The SDK supports 8 log severity levels with increasing severity:

- `debug` - Detailed debugging information
- `info` - General informational messages
- `notice` - Normal but significant events
- `warning` - Warning conditions
- `error` - Error conditions
- `critical` - Critical conditions
- `alert` - Action must be taken immediately
- `emergency` - System is unusable
```

dak2: That makes sense! Applied the changes.
https://github.com/modelcontextprotocol/ruby-sdk/compare/ac354523e320a9e5f39948bca653152437266119..44b969a31a3b4f126776791c2f2d4911510963b8

lib/mcp/server.rb
```diff
@@ -307,6 +322,19 @@ def init(request)
       }.compact
     end
 
+    def logging_level=(request)
```
koic: Hm, In Ruby, using a form like `logging_level = request`, where the attribute and the argument being assigned are not equivalent, feels unnatural. As an alternative, how about using a name like `configure_logging_level(request)`?

```suggestion
    def configure_logging_level(request)
```

dak2: I initially used `set_logging_level` to stay consistent with the [python-sdk](https://github.com/modelcontextprotocol/python-sdk/blob/65c614e48e833df4622b07da698ca64935c31f36/src/mcp/server/lowlevel/server.py#L419). 
However, since it was flagged by RuboCop's `Naming/AccessorMethodName` cop, I decided to go with `configure_logging_level` instead. 

https://github.com/modelcontextprotocol/ruby-sdk/compare/5b244e74622b3c09c472e7535a84104dce77f9ba..afa2fb40770c8c08e88a7c6587b9a3a389dacaa4

Thanks!

koic: Yeah, it is possible to disable the RuboCop rule and use `set_logging_level`, but the behavior expected from the name doesn't seem intuitive. In any case, since this is a private method, there is effectively no impact on users.

koic:
Let's take this as a starting point. Thank you!

dak2:
@koic 

I appreciate your continued support. Thank you!!

PR #107, opened 2025-08-11
Response error keyword argument

pusewicz:
The positional argument and naming is a bit awkward, and feels unnatural. Let's instead follow the Ruby style and add an `error:` keyword argument.

#### Before

```ruby
response = Response.new(content, true) # What is true?
```

#### After

```ruby
response = Response.new(content, error: true)
```

## How Has This Been Tested?

Unit tests

## Breaking Changes

For anyone using the positional argument, this is a breaking change.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

I'm happy to remove the second commit that introduces the `ErrorResponse` class. I do find it's usage a nicer approach that moves the decision to the caller, but also I can see how it adds some unnecessary boilerplate. Let me know what you think!

### DIFF
--- a/lib/mcp/tool/response.rb
+++ b/lib/mcp/tool/response.rb
@@ -3,15 +3,26 @@
 module MCP
   class Tool
     class Response
-      attr_reader :content, :is_error
+      NOT_GIVEN = Object.new.freeze
 
-      def initialize(content, is_error = false)
+      attr_reader :content
+
+      def initialize(content, deprecated_error = NOT_GIVEN, error: false)
+        if deprecated_error != NOT_GIVEN
+          warn("Passing `error` with the 2nd argument of `Response.new` is deprecated. Use keyword argument like `Response.new(content, error: error)` instead.", uplevel: 1)
+          error = deprecated_error
+        end
+
         @content = content
-        @is_error = is_error
+        @error = error
+      end
+
+      def error?
+        !!@error
       end
 
       def to_h
-        { content:, isError: is_error }.compact
+        { content:, isError: error? }.compact
       end
     end
   end
--- a/test/mcp/tool/response_test.rb
+++ b/test/mcp/tool/response_test.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module MCP
+  class Tool
+    class ResponseTest < ActiveSupport::TestCase
+      test "#initialize with content" do
+        content = [{
+          type: "text",
+          text: "Unauthorized",
+        }]
+        response = Response.new(content)
+
+        assert_equal content, response.content
+        refute response.error?
+      end
+
+      test "#initialize with content and error set to true" do
+        content = [{
+          type: "text",
+          text: "Unauthorized",
+        }]
+        response = Response.new(content, error: true)
+
+        assert_equal content, response.content
+        assert response.error?
+      end
+
+      test "#initialize with content and error explicitly set to false" do
+        content = [{
+          type: "text",
+          text: "Unauthorized",
+        }]
+        response = Response.new(content, error: false)
+
+        assert_equal content, response.content
+        refute response.error?
+      end
+
+      test "#error? for a standard response" do
+        response = Response.new(nil, error: false)
+        refute response.error?
+      end
+
+      test "#error? for an error response" do
+        response = Response.new(nil, error: true)
+        assert response.error?
+      end
+
+      test "#to_h for a standard response" do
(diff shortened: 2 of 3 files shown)


rtymchyk (approved):

lib/mcp/tool/error_response.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module MCP
+  class Tool
+    class ErrorResponse
+      def initialize(content) = super(content, error: true)
+    end
+  end
+end
```
koic: The introduction of this shorthand class seems redundant. `Response.new(content, error: true)` should be sufficient to express the intent.

lib/mcp/tool/response.rb
```diff
@@ -3,16 +3,18 @@
...
-        @is_error = is_error
+        @error = error
       end
 
       def to_h
-        { content:, isError: is_error }.compact
+        { content:, isError: error? }.compact
       end
+
+      alias_method :error?, :error
```
koic: It does not seem appropriate to expose `Response#error` as part of the interface. In other words, the only predicate interface that should be exposed is `Response#error?`.

Also, providing a migration path for the existing positional 2nd argument would be more user-friendly. This can be solved with a simple implementation.

These can be achieved with a diff like the following.

```diff
 module MCP
   class Tool
     class Response
-      attr_reader :content, :error
+      NOT_GIVEN = Object.new
+
+      attr_reader :content
+
+      def initialize(content, legacy_error = NOT_GIVEN, error: false)
+        if legacy_error != NOT_GIVEN
+          warn 'Passing `error` with the 2nd argument of `Response.new` is deprecated. Use keyword argument like `Response.new(content, error: error)` instead.', uplevel: 1
+          error = legacy_error
+        end
 
-      def initialize(content, error: false)
         @content = content
         @error = error
       end
 
+      def error?
+        !!@error
+      end
+
       def to_h
         { content:, isError: error? }.compact
       end
-
-      alias_method :error?, :error
     end
   end
 end
```

atesgoral: Yes, we'd like to avoid breaking changes as much as possible, since we're not using semantic versioning yet.

pusewicz: Done in https://github.com/modelcontextprotocol/ruby-sdk/pull/107/commits/10aa05556fee8a8b1325f98071ec1ab3a67e3fd2

lib/mcp/tool/error_response.rb
pusewicz: Dropped the commit.

test/mcp/tool/response_test.rb
```diff
@@ -0,0 +1,55 @@
...
+        assert_equal content, response.content
+        assert response.error?
+
+        response = Response.new(content, error: false)
+        assert_equal content, response.content
+        refute response.error?
+
+        response = Response.new(content)
+        assert_equal content, response.content
+        refute response.error?
```
koic, line 43: Can you split it into three tests, each independently executable per assertion group?

test/mcp/tool/response_test.rb
```diff
@@ -0,0 +1,55 @@
...
+
+        assert_equal [:content, :isError].sort, actual.keys.sort
+        assert_equal content, actual[:content]
+        refute actual[:isError]
+
+        response = Response.new(content, error: true)
+        actual = response.to_h
+        assert_equal [:content, :isError].sort, actual.keys.sort
+        assert_equal content, actual[:content]
+        assert actual[:isError]
```
koic, line 73: Can you split it into two tests, each independently executable per assertion group?

test/mcp/tool/response_test.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      test "#error?" do
+        response = Response.new(nil, error: true)
+        assert response.error?
+
+        response = Response.new(nil, error: false)
+        refute response.error?
```
koic: Can you split it into two tests, each independently executable per assertion?

test/mcp/tool/response_test.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
```
koic: Can you move this `test "#error?" do` (L46-L52) before `test "#to_h" do` (L27) ?

koic:
Since the purpose of these changes is singular, please squash them into a single commit.

pusewicz:
@koic Addressed your feedback.

Issue #111, opened 2025-08-21
Should the streamable HTTP transport respond with 202 Accepted instead of a 200?


stevegeek:
When testing in LM Studio an MCP server built with the `StreamableHTTPTransport`  I get an error about the body of the response to tool calls `"code": "unrecognized_keys", "keys": [ "accepted" ]"`

I tried modifying the response from `send_response_to_stream` (https://github.com/modelcontextprotocol/ruby-sdk/blob/f63f360a7b465fbb28a13f11ef82cfc0f40ca28b/lib/mcp/server/transports/streamable_http_transport.rb#L217) to a body-less 202 Accepted  and that fixes the tool call (and only parse the response if its not nil.. for some reason it seems to be nil for "notification_cancelled" method call)

ie

```ruby
        def send_response_to_stream(stream, response, session_id)
          if response
            message = JSON.parse(response)
            send_to_stream(stream, message)
          end
          [202, {}, []]
...
```

Looking at the spec I see https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server `If the server accepts the input, the server MUST return HTTP status code 202 Accepted with no body.` ... but Im not sure if thats the relavent section related to `send_response_to_stream` 

**Logs**

MCP Responded with
```
127.0.0.1 - - [21/Aug/2025:05:27:41 +0000] "POST / HTTP/1.1" 200 - 0.0007
[MCP] Request: notifications/initialized (id: )
[MCP] Request body: {
  "method": "notifications/initialized",
  "jsonrpc": "2.0"
}
127.0.0.1 - - [21/Aug/2025:05:27:41 +0000] "POST / HTTP/1.1" 202 - 0.0002
127.0.0.1 - - [21/Aug/2025:05:27:41 +0000] "GET / HTTP/1.1" 200 - 0.0002
[MCP] Request: tools/list (id: 1)
[MCP] Request body: {
  "method": "tools/list",
  "params": {},
  "jsonrpc": "2.0",
  "id": 1
}
[MCP] Response: empty (id: )
[MCP] Response body: {
  "accepted": true
}
127.0.0.1 - - [21/Aug/2025:05:27:41 +0000] "POST / HTTP/1.1" 200 - 0.0003
```

then LM Studio reports this parse error:

```
[...]

PR #114, opened 2025-08-24
Fix to return a response code 202 if the server accepts JSON-RPC notifications and responses

dak2:
## Motivation and Context

If the server accepts JSON-RPC notifications and responses, it must return an HTTP status code of 202 without a body.

refs
- https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server
- https://www.jsonrpc.org/specification

## How Has This Been Tested?

- [x] Check existing and added test cases for this case.
- [x] Confirm the server returns 202 when the client sends a notification or response.

## Breaking Changes

There is a possibility that, if users depend on status code 200.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Follow up: https://github.com/modelcontextprotocol/ruby-sdk/issues/111

### DIFF
--- a/lib/mcp/server/transports/streamable_http_transport.rb
+++ b/lib/mcp/server/transports/streamable_http_transport.rb
@@ -108,8 +108,8 @@ module MCP
 
           if body["method"] == "initialize"
             handle_initialization(body_string, body)
-          elsif body["method"] == MCP::Methods::NOTIFICATIONS_INITIALIZED
-            handle_notification_initialized
+          elsif notification?(body) || response?(body)
+            handle_accepted
           else
             handle_regular_request(body_string, session_id)
           end
@@ -168,6 +168,14 @@ module MCP
           [400, { "Content-Type" => "application/json" }, [{ error: "Invalid JSON" }.to_json]]
         end
 
+        def notification?(body)
+          !body["id"] && !!body["method"]
+        end
+
+        def response?(body)
+          !!body["id"] && !body["method"]
+        end
+
         def handle_initialization(body_string, body)
           session_id = SecureRandom.uuid
 
@@ -187,7 +195,7 @@ module MCP
           [200, headers, [response]]
         end
 
-        def handle_notification_initialized
+        def handle_accepted
           [202, {}, []]
         end
 
--- a/test/mcp/server/transports/streamable_http_transport_test.rb
+++ b/test/mcp/server/transports/streamable_http_transport_test.rb
@@ -620,6 +620,32 @@ module MCP
           assert_empty(response[2])
         end
 
+        test "handles POST request with body including JSON-RPC response object and returns with no body" do
+          init_request = create_rack_request(
+            "POST",
+            "/",
+            { "CONTENT_TYPE" => "application/json" },
+            { jsonrpc: "2.0", method: "initialize", id: "init" }.to_json,
+          )
+          init_response = @transport.handle_request(init_request)
+          session_id = init_response[1]["Mcp-Session-Id"]
+
+          request = create_rack_request(
+            "POST",
+            "/",
+            {
+              "CONTENT_TYPE" => "application/json",
+              "HTTP_MCP_SESSION_ID" => session_id,
+            },
+            { jsonrpc: "2.0", result: "success", id: "123" }.to_json,
+          )
+
+          response = @transport.handle_request(request)
+          assert_equal 202, response[0]
+          assert_empty(response[1])
+          assert_empty(response[2])
+        end
+
         private
 
         def create_rack_request(method, path, headers, body = nil)


atesgoral:
On the tests: I don't see anywhere in either the MCP or the JSON-RPC spec where a `result` or `error` property is used in a request with an id. These are response properties. Requests have `method` and `params`.

dak2:
@atesgoral 
Thanks for your review.

As you pointed out, it seems I misunderstood. I've deleted the test case I added.

Also, I believe this case is already covered by an existing test in `test/mcp/server/transports/streamable_http_transport_test.rb` (the case where a server returns 202 when a JSON-RPC notification request object is passed, and 200 when a JSON-RPC regular request object is passed), so I haven't added any specific tests.

Could you please review it again?

koic:
It seems odd that there is no reproduction test. Without a reproduction case for this issue, wouldn't it be difficult to notice if a regression occurs?

dak2:
@koic 

Thanks for your checking!

I believe the following three points need to be verified for this requirement:

1. When receiving a `JSON-RPC` notification object, return a 202 status code without a response body.
2. When receiving a `JSON-RPC` response object, return a 202 status code without a response body.
3. Do not return a 202 when receiving a regular `JSON-RPC` request object.

Regarding the second point, as atesgoral pointed out, since it's a response object, it doesn't get passed as a request in the first place, so I don't believe it needs to be verified in testing.

For case 1, see lines https://github.com/modelcontextprotocol/ruby-sdk/blob/eb0d9c0e6d671805aa668e2b14ef146c75c4c48d/test/mcp/server/transports/streamable_http_transport_test.rb#L596-L621, and case 3 is https://github.com/modelcontextprotocol/ruby-sdk/blob/eb0d9c0e6d671805aa668e2b14ef146c75c4c48d/test/mcp/server/transports/streamable_http_transport_test.rb#L20-L50

I believe these cases is already covered by above test cases, so I don't think we need to write additional tests. 
What do you think about this?

topherbullock:
> 4. If the input consists solely of (any number of) JSON-RPC responses or notifications:
> - If the server accepts the input, the server MUST return HTTP status code 202 Accepted with no body.

https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server

this doesn't exactly implement the spec, since only [`Notification` in JSON-RPC ](https://www.jsonrpc.org/specification#notification) is "a Request object without an "id" member". the ["Responses" type](https://www.jsonrpc.org/specification#response_object) includes an `id` field but does not have a method

instead of checking `!(body["id"] && body["method"])` , I'd prefer specific `response?(body)` `notification?(body)` checks to make the logic more clear that 202 is expected if input is only notification/response

dak2:
@topherbullock 

Thanks for the comment!!
I agree with you. It appears that my implementation logic was incorrect.
I separated the logic for the response and the notification.
I'd appreciate it if you could review it.

dak2:
@koic 

FYI: Sorry, I misunderstood. I've added a new test case for when the JSON-RPC request object is passed as input.

lib/mcp/server/transports/streamable_http_transport.rb
```diff
@@ -168,6 +168,14 @@ def parse_request_body(body_string)
           [400, { "Content-Type" => "application/json" }, [{ error: "Invalid JSON" }.to_json]]
         end
 
+        def notification?(body)
+          body["jsonrpc"] == "2.0" && !body["id"] && body["method"]
+        end
+
+        def response?(body)
+          body["jsonrpc"] == "2.0" && body["id"] && !body["method"]
```
koic: Wouldn't the condition `body["jsonrpc"] == "2.0"` be a validation that applies more broadly, rather than only to `notification?` and `response?`?

koic: Hm, doesn't this duplicate the validation in `json_rpc_handler`?
https://github.com/Shopify/json-rpc-handler/blob/v0.1.1/lib/json_rpc_handler.rb#L68

dak2: @koic 

Thank you for checking it!

As you pointed out, the `jsonrpc` validation is already handled by the `json_rpc_handler`, so it's unnecessary. I overlooked that.

Deleted.
https://github.com/modelcontextprotocol/ruby-sdk/compare/9631b57dd374c74ed67809ac9e962d1fa84c6c62..2aee943b1067f3105498ae3ab0e51b83f7bc1eb6

lib/mcp/server/transports/streamable_http_transport.rb
```diff
@@ -168,6 +168,14 @@ def parse_request_body(body_string)
           [400, { "Content-Type" => "application/json" }, [{ error: "Invalid JSON" }.to_json]]
         end
 
+        def notification?(body)
+          !body["id"] && body["method"]
```
koic: Nits, the values of `body["method"]` doesn't return to be a boolean. Since these is a predicate method, it would be better to return a boolean using something like `!!`.

dak2: I made the change you suggested because I thought it was a more accurate approach.
https://github.com/modelcontextprotocol/ruby-sdk/compare/2aee943b1067f3105498ae3ab0e51b83f7bc1eb6..cd7a54a5c9110035a530dddb1607329ddf4e78a9

Thank you for your review!

atesgoral (approved):

koic:
Thanks!

Issue #142, opened 2025-09-23
Output schema for arrays


nilcolor:
**Describe the bug**
When output schema is described like
```
{
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "name": {"type": "string"},
      "nick": {"type": "string"},
    },
    "required": ["name", "nick"]
  }
}
```

MCP clients will see just (`{type: "object}`) because of this logic https://github.com/modelcontextprotocol/ruby-sdk/blob/main/lib/mcp/tool/output_schema.rb#L22-L27 (Am I right?).

**To Reproduce**
1. Create a new tool like
```
class TestTool < MCP::Tool
  output_schema({
    description: "A list of things",
    type: "array",
    items: {
      type: "object",
      properties: {
        name: {type: "string"},
        nick: {type: "string"},
      },
      required: ["name", "nick"],
    },
  })
end
```
2. Check tool's information using something like `@modelcontextprotocol/inspector`

**Expected behavior**
It is expected that tool can notify client about output schema that is a list of something.

**Logs**
none

**Additional context**
I can try to propose a PR... Might not cover all the cases tho. Unless `{ type: "object" }.tap` was used specifically and is not subject to change,

PR #144, opened 2025-09-24
Allow output schema to be array of objects

nilcolor:
This change allows you to declare the output schema as an array of objects.
When building an MCP, you might want to return a list of results while still providing a schema for those results. This helps clients better understand the data being returned.

## Motivation and Context
One of the MCPs I'm building need to respond with list of items. And I'd like to provide a schema for this. Currently this is not possible (#142).

## How Has This Been Tested?
I tried to test in `@modelcontextprotocol/inspector` but Inspector doesn't support it yet (https://github.com/modelcontextprotocol/inspector/issues/552)
In real application it seems to work fine.

## Breaking Changes
OutputSchema class no longer expose `#properties`, `#required`. But most likely they were not used by any consumer. Other than this, no breaking changes and user do not need to update existing configurations.

## Types of changes

- [ ] Documentation update

## Checklist


README.md
```diff
@@ -546,6 +546,26 @@ class DataTool < MCP::Tool
...
+        temperature: { type: "number" },
+        condition: { type: "string" },
+        humidity: { type: "integer" }
+      },
+      required: ["temperature", "condition", "humidity"]
+    }
+  )
+end
+```
+Please note: in this case, you must provide `type: "array"`. The default type
```
koic, line 567: Please insert a blank line.

```suggestion

Please note: in this case, you must provide `type: "array"`. The default type
```

lib/mcp/tool/output_schema.rb
```diff
@@ -1,29 +1,27 @@
 # frozen_string_literal: true
 
+require "json"
```
koic: This `require "json"` seems unnecessary.

```suggestion
```

lib/mcp/tool/output_schema.rb
```diff
@@ -35,8 +33,23 @@ def validate_result(result)
 
       private
 
+      def deep_transform_keys(hash, &block)
+        case hash
+        when Hash
+          hash.each_with_object({}) do |(key, value), result|
+            if key.casecmp?("$ref") || key == :$ref
```
koic: Are there cases with symbols? If not, it seems safe to remove.

```suggestion
            if key.casecmp?("$ref")
```

lib/mcp/tool/output_schema.rb
```diff
@@ -35,8 +33,23 @@ def validate_result(result)
 
       private
 
+      def deep_transform_keys(hash, &block)
```
koic: `schema` or `json_schema` might be more in line with the intent:

```suggestion
      def deep_transform_keys(schema, &block)
```

lib/mcp/tool/output_schema.rb
nilcolor: I'm OK to remove. Added because original check was compare with strings and symbols: `if obj.key?("$ref") || obj.key?(:$ref)`

nilcolor:
@koic Thank you. I've updated PR with your suggestions.

koic:
@nilcolor This looks good to me. Can you squash your commits into one?

nilcolor:
@koic Done

PR #149, opened 2025-10-01
[client] Change call_tool method to return the full tool response to help get structuredContent

juharris:
⚠️ Change `call_tool` in `Client` to return the full response.

## Motivation and Context
Helps get `structuredContent` and errors.

More like Python does:
https://github.com/modelcontextprotocol/python-sdk/blob/3e798bfc01856df711b921f7f3b55379fbd087e4/src/mcp/client/session.py#L279

## How Has This Been Tested?
Unit test

## Breaking Changes
⚠️ YES!

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/client.rb
+++ b/lib/mcp/client.rb
@@ -44,28 +44,27 @@ module MCP
       end || []
     end
 
-    # Calls a tool via the transport layer.
+    # Calls a tool via the transport layer and returns the full response from the server.
     #
     # @param tool [MCP::Client::Tool] The tool to be called.
     # @param arguments [Object, nil] The arguments to pass to the tool.
-    # @return [Object] The result of the tool call, as returned by the transport.
+    # @return [Hash] The full JSON-RPC response from the transport.
     #
     # @example
     #   tool = client.tools.first
-    #   result = client.call_tool(tool: tool, arguments: { foo: "bar" })
+    #   response = client.call_tool(tool: tool, arguments: { foo: "bar" })
+    #   structured_content = response.dig("result", "structuredContent")
     #
     # @note
     #   The exact requirements for `arguments` are determined by the transport layer in use.
     #   Consult the documentation for your transport (e.g., MCP::Client::HTTP) for details.
     def call_tool(tool:, arguments: nil)
-      response = transport.send_request(request: {
+      transport.send_request(request: {
         jsonrpc: JsonRpcHandler::Version::V2_0,
         id: request_id,
         method: "tools/call",
         params: { name: tool.name, arguments: arguments },
       })
-
-      response.dig("result", "content")
     end
 
     private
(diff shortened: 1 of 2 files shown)


lib/mcp/client.rb
```diff
@@ -58,14 +58,32 @@ def tools
...
+    #
+    # @example
+    #   tool = client.tools.first
+    #   response = client.call_tool_with_response(tool: tool, arguments: { foo: "bar" })
+    #   structured_content = response.dig("result", "structuredContent")
+    #
+    # @note
+    #   The exact requirements for `arguments` are determined by the transport layer in use.
+    #   Consult the documentation for your transport (e.g., MCP::Client::HTTP) for details.
+    def call_tool_raw(tool:, arguments: nil)
```
juharris: This method name, `call_tool_raw` doesn't thrill me. Any other ideas?

lib/mcp/client.rb
```diff
@@ -58,14 +58,32 @@ def tools
...
+      transport.send_request(request: {
         jsonrpc: JsonRpcHandler::Version::V2_0,
         id: request_id,
         method: "tools/call",
         params: { name: tool.name, arguments: arguments },
       })
-
-      response.dig("result", "content")
```
atesgoral, line 68: Hmm. I hadn't paid attention to this. Perhaps the client contract should have been to return the entire result object. We're also losing `isError` with this. A better fix to losing `structuredContent` might be to make a breaking change to the client to return the whole result.

juharris: I sort of pointed it out, but wasn't specific enough: https://github.com/modelcontextprotocol/ruby-sdk/pull/28#discussion_r2239851160

I'm okay with breaking changes, but it would be hard to notice issues as the method would just return a different "hash", which is not clear. Perhaps we should change the method name and make the new one look like the proper one:
`call_server_tool`?

topherbullock: I am fine with breaking changes given we are pre-1.0.0, as long as we call out the breaking change in the release notes.

topherbullock: > `@return [Hash] The full JSON-RPC response from the transport.`

the documentation is not accurate for  `#call_tool`, and I would prefer breaking changes that make the documentation true rather than keeping backward compatibility of existing  `call_tool` behaviour

juharris: BTW Looks like Python's `call_tool` returns the full result: https://github.com/modelcontextprotocol/python-sdk/blob/3e798bfc01856df711b921f7f3b55379fbd087e4/src/mcp/client/session.py#L279

Let's do that too.

juharris: I changed `call_tool` and now this is a breaking change.

topherbullock (approved):

PR #150, opened 2025-10-02
Set default `content` to an empty array instead of `nil`

koic:
## Motivation and Context

Follow-up to https://github.com/modelcontextprotocol/ruby-sdk/pull/147.

According to the specification schema, `content` is not optional and therefore cannot be omitted:

```typescript
interface CallToolResult {
  _meta?: { [key: string]: unknown };
  content: ContentBlock[];
  isError?: boolean;
  structuredContent?: { [key: string]: unknown };
  [key: string]: unknown;
}
```

https://modelcontextprotocol.io/specification/2025-06-18/schema#calltoolresult

Instead of `nil`, an empty array is set as the default value.

There may be a better value for `content`, but at the very least it should not be missing.

## How Has This Been Tested?

The related test code has been updated.

## Breaking Changes

#147 is unreleased, there are no breaking changes on the main branch.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/tool/response.rb
+++ b/lib/mcp/tool/response.rb
@@ -13,7 +13,7 @@ module MCP
           error = deprecated_error
         end
 
-        @content = content
+        @content = content || []
         @error = error
         @structured_content = structured_content
       end
--- a/test/mcp/tool/response_test.rb
+++ b/test/mcp/tool/response_test.rb
@@ -101,13 +101,24 @@ module MCP
         refute actual[:isError]
       end
 
+      test "#to_h for a standard response with nil content and structured content" do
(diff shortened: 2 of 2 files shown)


atesgoral (approved):

lib/mcp/tool/response.rb
juharris: Should we instead handle when `nil` is passed to this method?
```suggestion
        @content = content || []
```

koic: Yeah, that makes sense! I've added test code for that case and updated this PR.

juharris (approved):

atesgoral (approved):

PR #163, opened 2025-10-11
Add `prompts/list` and `prompts/get` support to client

dak2:
## Motivation and Context

This commit adds two new methods to `MCP::Client`:
- prompts: Lists available prompts from the server via `prompts/list`
- get_prompt: Retrieves a specific prompt by name via `prompts/get`

Both methods follow the same pattern as [the existing resources support](https://github.com/modelcontextprotocol/ruby-sdk/pull/160), including comprehensive test coverage and README documentation updates.

ref. https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/2025-03-26/server/prompts.mdx#protocol-messages

## How Has This Been Tested?

Check existing and added test cases for this case.
https://github.com/modelcontextprotocol/ruby-sdk/pull/163/files#diff-7daf60c70888017747b710518764087efc6171bb2f5419506cf21a8ac7ef4a5a

## Breaking Changes

None

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/client.rb
+++ b/lib/mcp/client.rb
@@ -58,6 +58,20 @@ module MCP
       response.dig("result", "resources") || []
     end
 
+    # Returns the list of prompts available from the server.
+    # Each call will make a new request – the result is not cached.
+    #
+    # @return [Array<Hash>] An array of available prompts.
+    def prompts
+      response = transport.send_request(request: {
+        jsonrpc: JsonRpcHandler::Version::V2_0,
+        id: request_id,
+        method: "prompts/list",
+      })
+
+      response.dig("result", "prompts") || []
+    end
+
     # Calls a tool via the transport layer and returns the full response from the server.
     #
     # @param tool [MCP::Client::Tool] The tool to be called.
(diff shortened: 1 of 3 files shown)


lib/mcp/client.rb
```diff
@@ -96,6 +110,21 @@ def read_resource(uri:)
...
+    # @return [Hash] A hash containing the prompt details.
+    def get_prompt(name:)
+      response = transport.send_request(request: {
+        jsonrpc: JsonRpcHandler::Version::V2_0,
+        id: request_id,
+        method: "prompts/get",
+        params: { name: name },
+      })
+
+      response.dig("result") || {}
```
koic: It seems there's no test for the default value. Also, `fetch` might be more suitable in this kind of case.

```suggestion
      response.fetch("result", {})
```

dak2: @koic 

Thanks for your review!

I added test cases for this scenario and switched to using `fetch`.
https://github.com/modelcontextprotocol/ruby-sdk/compare/67225d7e1400f14b43c996bc165a5059df686d14..e88deb6b65beebf3a28f3800d7e5008b87636507

PR #165, opened 2025-10-11
Fix tool error handling to follow MCP spec

### LINKED ISSUE #159: Tool error handling does not follow spec
**Describe the bug**
According to the [docs](https://github.com/modelcontextprotocol/ruby-sdk?tab=readme-ov-file#exception-reporting) (and the relevant portion of the [MCP spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling)), tool execution errors should look like the following:

```
{
  "jsonrpc": "2.0",
  "id": 4,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "Failed to fetch weather data: API rate limit exceeded"
      }
    ],
    "isError": true
  }
}
```

But instead, they look like the following (i.e., protocol error format):

```
{
  "jsonrpc": "2.0",
  "id": 4,
  "error": {
    "code": -32603,
    "message": "Internal error",
    "data": "Internal error calling tool fetch_weather_data"
  }
}
```

So the **actual behavior** of the MCP gem stack is:
1. Tool raises an exception (e.g., `ActiveRecord::RecordNotFound`)
2. [MCP Server's `call_tool` catches it and re-raises as `RequestHandlerError`](https://github.com/modelcontextprotocol/ruby-sdk/blob/77370cbb93fe7b05a9c6cade043064b8e807f4c1/lib/mcp/server.rb#L291)
3. Seemingly, it is caught and re-raised [here](https://github.com/modelcontextprotocol/ruby-sdk/blob/77370cbb93fe7b05a9c6cade043064b8e807f4c1/lib/mcp/server.rb#L221) and somewhere `JsonRpcHandler` catches that and returns a JSON-RPC error response (had a bit of trouble finding exactly where the caught error lands).

Unless I'm missing something, this means the MCP gem documentation is incorrect - it does not automatically convert exceptions to `isError: true` responses. Instead, raising exceptions get converted to JSON-RPC protocol errors (code -32603). 

Returning `Tool::Response.new(content, error: true)` would work as documented, but requires manual exception handling as a workaround.

**To Reproduce**
1. Raise an error inside a tool call.
2. Verify that response matches what the MCP spec deems a protocol error rather than a tool execution error.

**Expected behavior**
I expect for raised errors within tool calls (e.g., `ActiveRecord::RecordNotFound`) that are not deemed internal server errors to be handled intuitively, as tool call execution errors, rather than protocol (-32603) errors. This is what the README says should be the current default behavior for raised errors within tool calls.

Jflick58:
Tool execution errors now return successful responses with isError: true
  instead of JSON-RPC protocol errors, per the MCP specification.

## Motivation and Context
This allows clients to distinguish between protocol-level errors and
tool execution errors, providing better error context.

Fixes #159

## How Has This Been Tested?
Updated unit tests to support new error behavior. All existing tests passing. 

## Breaking Changes
If client libs are parsing the error message for the specific tool call error, then they will likely get a parser error due to the change of the error structure. 

Current behavior:
 ```
     {
       "jsonrpc": "2.0",
       "id": 4,
       "error": {
         "code": -32603,
         "message": "Internal error",
         "data": "Internal error calling tool fetch_weather_data"
       }
     }
```

Updated behavior (matches spec and readme):

```
     {
       "jsonrpc": "2.0",
       "id": 4,
       "result": {
         "content": [{
           "type": "text",
           "text": "Internal error calling tool fetch_weather_data"
         }],
         "isError": true
       }
     }
```

## Types of changes
- [ ] Documentation update

## Checklist
- [x ] I have added or updated documentation as needed

### DIFF
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -258,10 +258,17 @@ module MCP
 
     def call_tool(request)
       tool_name = request[:name]
+
       tool = tools[tool_name]
       unless tool
-        add_instrumentation_data(error: :tool_not_found)
-        raise RequestHandlerError.new("Tool not found #{tool_name}", request, error_type: :tool_not_found)
+        add_instrumentation_data(tool_name:, error: :tool_not_found)
+        return Tool::Response.new(
+          [{
+            type: "text",
+            text: "Tool not found: #{tool_name}",
+          }],
+          error: true,
+        ).to_h
       end
 
       arguments = request[:arguments] || {}
@@ -269,11 +276,14 @@ module MCP
 
       if tool.input_schema&.missing_required_arguments?(arguments)
         add_instrumentation_data(error: :missing_required_arguments)
-        raise RequestHandlerError.new(
-          "Missing required arguments: #{tool.input_schema.missing_required_arguments(arguments).join(", ")}",
-          request,
-          error_type: :missing_required_arguments,
-        )
+        missing = tool.input_schema.missing_required_arguments(arguments).join(", ")
+        return Tool::Response.new(
+          [{
+            type: "text",
+            text: "Missing required arguments: #{missing}",
+          }],
+          error: true,
+        ).to_h
       end
 
       if configuration.validate_tool_call_arguments && tool.input_schema
@@ -281,14 +291,27 @@ module MCP
           tool.input_schema.validate_arguments(arguments)
         rescue Tool::InputSchema::ValidationError => e
           add_instrumentation_data(error: :invalid_schema)
-          raise RequestHandlerError.new(e.message, request, error_type: :invalid_schema)
+          return Tool::Response.new(
+            [{
+              type: "text",
+              text: e.message,
+            }],
+            error: true,
+          ).to_h
         end
       end
 
       begin
         call_tool_with_args(tool, arguments)
       rescue => e
-        raise RequestHandlerError.new("Internal error calling tool #{tool_name}", request, original_error: e)
+        report_exception(e, { request: request })
+        Tool::Response.new(
+          [{
+            type: "text",
+            text: "Internal error calling tool #{tool_name}: #{e.message}",
+          }],
+          error: true,
+        ).to_h
       end
     end
 
--- a/test/mcp/server_context_test.rb
+++ b/test/mcp/server_context_test.rb
@@ -152,9 +152,10 @@ module MCP
 
       response = server_no_context.handle(request)
 
-      assert response[:error]
-      # The error is wrapped as "Internal error calling tool..."
-      assert_equal "Internal error", response[:error][:message]
+      assert_nil response[:error], "Expected no JSON-RPC error"
+      assert response[:result][:isError]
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: lib/mcp/server.rb [hunk 1]
```diff
@@ -288,7 +288,14 @@ def call_tool(request)
       begin
         call_tool_with_args(tool, arguments)
       rescue => e
```
atesgoral, line 306: This is a step in the right direction, but I think we can take this a few more steps further:

This rescue should be at the `call_tool` level: Any mistake made by the caller that falls out of valid tool contracts (e.g. invalid tool name, invalid args) should be non-exception tool results with `isError: true`. Only JSON-RPC-level protocol errors should be allowed to be returned as JSON-RPC errors with error codes.

Jflick58: @atesgoral  makes sense. This is how it should work now: 

  Original (Everything was a Protocol Error)
```
  {
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
      "code": -32603,
      "message": "Internal error",
      "data": "Tool not found unknown_tool"
    }
  }
```
  Now - Protocol Error 
```
  {
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
      "code": -32601,
      "message": "Method not found",
      "data": "unknown_method"
    }
  }
```

 Now - Tool error: 
```
  {
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
      "content": [{
        "type": "text",
        "text": "Tool not found: unknown_tool"
      }],
      "isError": true
    }
  }
```

lib/mcp/server.rb
```diff
@@ -258,37 +258,60 @@ def list_tools(request)
 
     def call_tool(request)
       tool_name = request[:name]
+      arguments = request[:arguments] || {}
```
koic: It seems better to keep it at its original position on line 267, since there's no need to process it until it becomes necessary.

koic:
I've left one small comment, but overall it looks good to me. Can you address it and squash the commits in to one?

koic:
@Jflick58 ping.

lib/mcp/server.rb
```diff
@@ -258,37 +258,60 @@ def list_tools(request)
...
 
       begin
         call_tool_with_args(tool, arguments)
       rescue => e
-        raise RequestHandlerError.new("Internal error calling tool #{tool_name}", request, original_error: e)
+        report_exception(e, { request: request })
+        Tool::Response.new(
+          [{
+            type: "text",
+            text: "Internal error calling tool #{tool_name}",
```
cdelafuente-r7: I think it would be useful to also include any details the tool reported when raising this exception. For example:
```suggestion
            text: "Internal error calling tool #{tool_name}: #{e.message}",
```

Jflick58:
@koic  sorry have been sick. I'll get this buttoned up

koic (approved):
@Jflick58 Thank you for working on this!

PR #175, opened 2025-10-30
Internalize json_rpc_handler along with an XSS vulnerability fix

atesgoral:
## Motivation and Context

The ruby-sdk gem is the only popular gem that uses the json_rpc_handler gem. For tweaks around transport error handling, there is an ongoing awkwardness of working around the opinions of the json_rpc_handler gem or needing to extend that gem to make its API more flexible.

Just internalizing the JSON-RPC handling will alllow mcp gem development to move faster.

Incidentally, we also pick up an unreleased security fix on json_rpc_handler:
- [x] https://github.com/Shopify/json-rpc-handler/pull/23

**Reviewers:** Follow individual commits to see the incremental steps I took in copying things over, so you're not tasked with reviewing the entire internalized "new" code. Code copied from commit https://github.com/Shopify/json-rpc-handler/commit/b41c41249f5e68fabb65148d5434d3349026a7ab

## How Has This Been Tested?
Existing unit tests pass.

## Breaking Changes
The gem becomes stringent on the format of JSON-RPC call ids by default, to prevent an XSS vulnerability.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/lib/mcp.rb
+++ b/lib/mcp.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative "json_rpc_handler"
 require_relative "mcp/configuration"
 require_relative "mcp/content"
 require_relative "mcp/instrumentation"
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "json_rpc_handler"
+require_relative "../json_rpc_handler"
 require_relative "instrumentation"
 require_relative "methods"
 
--- a/mcp.gemspec
+++ b/mcp.gemspec
@@ -29,6 +29,5 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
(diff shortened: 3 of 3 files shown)


lib/json_rpc_handler.rb
atesgoral, line 1: I preserved the name, but this could be moved to a nicer, `json-rpc` folder to follow suit with `mcp`.

lib/json_rpc_handler.rb
```diff
@@ -0,0 +1,151 @@
...
+
+  class ErrorCode
+    INVALID_REQUEST = -32600
+    METHOD_NOT_FOUND = -32601
+    INVALID_PARAMS = -32602
+    INTERNAL_ERROR = -32603
+    PARSE_ERROR = -32700
+  end
+
+  DEFAULT_ALLOWED_ID_CHARACTERS = /\A[a-zA-Z0-9_-]+\z/
```
atesgoral, line 19: As a follow up, we need to decide whether we want to keep this as configurable through the mcp gem, or allow developers to punch through layers to be able to directly override it at JSON-RPC module level, or simply make the mcp gem super opinionated about always having strict id validation. Heuristically this pattern should cover any non-pen-tester scenarios.

koic:
Thank you!

PR #199, opened 2025-12-12
Raise an error when duplicate tool names are registered

### LINKED ISSUE #197: Tools with the same name in different modules override each other
**Describe the bug**

Tools sharing the same name across different modules aren’t working, likely because of this line: https://github.com/modelcontextprotocol/ruby-sdk/blob/main/lib/mcp/server.rb#L54

**To Reproduce**
If your tool structure looks like this, for example

```
app/tools
├── application_tool.rb
├── clients
│   ├── create_tool.rb
│   ├── destroy_tool.rb
│   ├── index_tool.rb
│   ├── show_tool.rb
│   └── update_tool.rb
├── projects
│   ├── create_tool.rb
│   ├── destroy_tool.rb
│   ├── index_tool.rb
│   ├── show_tool.rb
│   └── update_tool.rb
```

Only the most recently registered tool remains in the tools list:

```ruby
[Clients::UpdateTool, Projects::UpdateTool].to_h { |t| [t.name_value, t] }
# => {"update_tool" => Projects::UpdateTool}
```

This happens because `name_value` returns `update_tool` for both classes.

**Expected behavior**
We should probably use a more robust key here to properly handle this case. I could open a PR if needed and adapt the tests, but let me know first what you think about it! 

Thanks for this gem! 🙌

koic:
## Motivation and Context

This raises an exception when duplicate tool names are registered, instead of silently overwriting tools. Tool names are required to be unique within a server, so this behavior aligns with the MCP specification.

> Tool names SHOULD be unique within a server.

https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-names

Validation for tool names could be made more strictly spec-compliant in a separate effort.

Fixes #197

## How Has This Been Tested?

Regression and updating tests pass.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -5,6 +5,15 @@ require_relative "instrumentation"
 require_relative "methods"
 
 module MCP
+  class ToolNotUnique < StandardError
+    def initialize(duplicated_tool_names)
+      super(<<~MESSAGE)
+        Tool names should be unique. Use `tool_name` to assign unique names to:
+        #{duplicated_tool_names.join(", ")}
+      MESSAGE
+    end
+  end
+
   class Server
     DEFAULT_VERSION = "0.1.0"
 
@@ -53,6 +62,7 @@ module MCP
       @title = title
       @version = version
       @instructions = instructions
+      @tool_names = tools.map(&:name_value)
       @tools = tools.to_h { |t| [t.name_value, t] }
       @prompts = prompts.to_h { |p| [p.name_value, p] }
       @resources = resources
@@ -101,7 +111,10 @@ module MCP
 
     def define_tool(name: nil, title: nil, description: nil, input_schema: nil, annotations: nil, meta: nil, &block)
       tool = Tool.define(name:, title:, description:, input_schema:, annotations:, meta:, &block)
-      @tools[tool.name_value] = tool
+      tool_name = tool.name_value
+
+      @tool_names << tool_name
+      @tools[tool_name] = tool
 
       validate!
     end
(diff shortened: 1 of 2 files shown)


atesgoral:
Won't this make the default tool names overly verbose, as they're presented to the LLM? Can't developers use the existing escape hatch of defining a custom name when registering a tool?

I also don't see a single MCP server/controller exposing tools with the same names as being a common use case. If the use case is an MCP gateway, then it could be the responsibility of the gateway (acting as host AND server) to apply a convention for prefixing tool names from upstream MCP servers.

koic:
Thank you for the well-reasoned feedback.

I do have concerns about tool names becoming longer. However, the effect applies only to default tool names when no explicit name such as `tool_name: "custom name"` is provided. Using a fully qualified name as the default could be seen as a clear and unambiguous representation.

As reported in #197, current default tool naming can cause tools to be unexpectedly dropped, and that outcome is concerning from a user-experience standpoint.

If a tool name that is more suitable for an LLM (or simply shorter and more concise) is preferred, it should be advisable to specify one explicitly with `tool_name: "more preferable name"`.

There is also a possibility of providing a warning from MCP server to the LLM when a configuration such as `tool_name: "duplicated name"` results in duplicate names, although that would be a separate feature to consider.

atesgoral:
How about we raise an exception when a duplicate tool name is registered? So the developer would know and override the name. Anything we do to prevent the silent ignoring of tools is better here.

koic:
> How about we raise an exception when a duplicate tool name is registered? 

Yeah, this does seem like a good approach. Since "`Tool names **SHOULD** be unique within a server`", it should also align with the MCP specification.
https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-names

After this, I will update the PR with this approach!

koic:
I've updated this PR to raise an error when duplicate tool names are registered.

NOTE: Validation for tool names could be made more strictly spec-compliant in a separate effort.

PR #206, opened 2026-01-05
Support Ruby 2.7 - Ruby 3.1

koic:
## Motivation and Context

Unlike applications, libraries provide value by remaining usable for a broad range of users, even when that means supporting older Ruby versions beyond Ruby's own EOL timeline.

I maintain RuboCop, a Ruby linter. Because linters need to work across the ecosystem, they often need to keep compatibility with older Ruby versions to some extent.

The trade-off is that the MCP Ruby SDK must remain compatible with Ruby 2.7. However, given the ecosystem impact, supporting Ruby 2.7 is still worthwhile.

For example, bundled gems such as `csv` (Ruby >= 2.5), `bigdecimal` (Ruby >= 2.5), `json` (Ruby >= 2.7), and `language_server-protocol` (Ruby >= 2.5) tend to keep compatibility with older Ruby versions.

## How Has This Been Tested?

`rubocop-shopify` (2.18) supports Ruby 3.1+, and this PR does not change that. Since RuboCop can configure the runtime Ruby version and the target Ruby version for analysis independently, CI continues to run RuboCop on Ruby 4.0, while analyzing code with `TargetRubyVersion: 2.7`.

When setting `TargetRubyVersion: 2.7`, several RuboCop offenses were resolved.

Additionally, Sorbet-related libraries support Ruby 3.0+. Since these are not dependencies of the MCP Ruby SDK itself, they are versioned and tested in the development Gemfile.

New test matrix for up to Ruby 2.7 has been added to CI.

## Breaking Changes

Unlike dropping support, which imposes restrictions on users, this change broadens compatibility and does not introduce breaking changes.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ jobs:
     strategy:
       matrix:
         entry:
+          - { ruby: '2.7', allowed-failure: false }
+          - { ruby: '3.0', allowed-failure: false }
+          - { ruby: '3.1', allowed-failure: false }
           - { ruby: '3.2', allowed-failure: false }
           - { ruby: '3.3', allowed-failure: false }
           - { ruby: '3.4', allowed-failure: false }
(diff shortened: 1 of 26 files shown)


atesgoral:
Outside of the context of this PR: I don't think we even need to stick to `rubocop-shopify` but switch to `rubocop` if that's going to be closer to community standards, lead to more poetic Ruby, etc.

gabrielrdrguez:
Thank you for this one, I was doing this myself today 'cause I needed it to run on 2.7 but you were ahead of me

PR #213, opened 2026-01-12
chore: update licensing to Apache 2.0 for new contributions

domdomegg:
As part of MCP joining the Linux Foundation, this PR updates licensing to Apache 2.0 for new contributions while grandfathering existing MIT code. See modelcontextprotocol/modelcontextprotocol#1994


koic:
AFAIK, the MIT License is indeed more commonly used than the Apache License 2.0 in the Ruby / Rails ecosystem. That said, this is something that should be addressed as part of MCP's overall governance.

From an operational perspective, a dual-licensing approach can also be considered. This approach can be understood as an alternative solution in the form of a "License in License".

Thank you!

PR #221, opened 2026-01-28
Add client info to instrumentation callback

duboff:
## Motivation and Context
Adding client information to the instrumentation callback. A follow-up to https://github.com/modelcontextprotocol/ruby-sdk/pull/218 of sorts. Client info is very helpful for analytics and debugging purposes.

## How Has This Been Tested?
Yes, tested with the Chattermill MCP (proprietary).

## Breaking Changes
No breaking changes 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x ] My code follows the repository's style guidelines
- [ x] New and existing tests pass locally
- [ x] I have added appropriate error handling
- [ x] I have added or updated documentation as needed

PR #223, opened 2026-02-01
Add protocol version negotiation to server initialization

duboff:
## Motivation and Context
Currently server launched enforces the latest protocol regardless of the protocols supported by the client. This leads to issues when a client (eg. Claude Desktop) supports only protocol versions (2025-06-18) behind the current version supported by the server (2025-11-25).

This PR adds full protocol version negotiation logic (as per the [MCP Spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#version-negotiation)), including dropping parameters not supported by each individual protocol version, to maximize compatibility.

## How Has This Been Tested?
Tested with the Chattermill MCP (proprietary). 

## Breaking Changes
No breaking changes expected.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
A potential consideration could be to add a config option to enumerate supported protocols if the server developer wants to explicitly avoid supporting all possible protocols. Not implemented here as doesn't seem necessary.

### DIFF
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -18,6 +18,9 @@ module MCP
   class Server
     DEFAULT_VERSION = "0.1.0"
 
+    UNSUPPORTED_PROPERTIES_UNTIL_2025_06_18 = [:description, :icons].freeze
+    UNSUPPORTED_PROPERTIES_UNTIL_2025_03_26 = [:title, :websiteUrl].freeze
+
     class RequestHandlerError < StandardError
       attr_reader :error_type
       attr_reader :original_error
@@ -319,13 +322,32 @@ module MCP
       }.compact
     end
 
-    def init(request)
-      @client = request[:clientInfo] if request
+    def init(params)
+      @client = params[:clientInfo] if params
+
+      protocol_version = params[:protocolVersion] if params
+      negotiated_version = if Configuration::SUPPORTED_STABLE_PROTOCOL_VERSIONS.include?(protocol_version)
+        protocol_version
+      else
+        configuration.protocol_version
+      end
+
+      info = server_info.reject do |property|
+        negotiated_version <= "2025-06-18" && UNSUPPORTED_PROPERTIES_UNTIL_2025_06_18.include?(property) ||
+          negotiated_version <= "2025-03-26" && UNSUPPORTED_PROPERTIES_UNTIL_2025_03_26.include?(property)
+      end
+
+      response_instructions = instructions
+
+      if negotiated_version == "2024-11-05"
+        response_instructions = nil
+      end
+
       {
-        protocolVersion: configuration.protocol_version,
+        protocolVersion: negotiated_version,
         capabilities: capabilities,
-        serverInfo: server_info,
-        instructions: instructions,
+        serverInfo: info,
+        instructions: response_instructions,
       }.compact
     end
 
--- a/test/mcp/server_test.rb
+++ b/test/mcp/server_test.rb
@@ -1274,6 +1274,95 @@ module MCP
       assert_equal custom_version, response[:result][:protocolVersion]
     end
 
+    test "server negotiates protocol version when client requests a supported version" do
+      server = Server.new(name: "test_server")
+
+      request = {
+        jsonrpc: "2.0",
+        method: "initialize",
+        id: 1,
+        params: {
+          protocolVersion: "2025-06-18",
+        },
+      }
+
+      response = server.handle(request)
+      assert_equal "2025-06-18", response[:result][:protocolVersion]
+    end
+
+    test "server falls back to default version when client requests unsupported version" do
+      server = Server.new(name: "test_server")
+
+      request = {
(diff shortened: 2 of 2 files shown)


lib/mcp/server.rb
```diff
@@ -313,12 +313,36 @@ def server_info
       }.compact
     end
 
-    def init(request)
+    def init(params)
+      client_version = params&.dig(:protocolVersion) || params&.dig("protocolVersion")
+      negotiated_version = if client_version && Configuration::SUPPORTED_STABLE_PROTOCOL_VERSIONS.include?(client_version)
+        client_version
```
koic: Shouldn't this be `protocol_version` rather than `client_version`? Also, it's not clear whether a string key is necessary here. For example, see https://github.com/modelcontextprotocol/ruby-sdk/pull/221/files#r2759263967.

```suggestion
      protocol_version = params[:protocolVersion] if params
      negotiated_version = if protocol_version && Configuration::SUPPORTED_STABLE_PROTOCOL_VERSIONS.include?(protocol_version)
        protocol_version
```


Additionally, should `request` be changed to `params`?

duboff: Pushed a change addressing most comments. I do think params makes more sense given how the method is used [here](https://github.com/modelcontextprotocol/ruby-sdk/blob/0ab77c318f91e36bd3db7d27805cf59b500df1ba/lib/mcp/server.rb#L281) but obviously it doesn't really matter so happy to change if you prefer to keep it the same.

atesgoral (approved):
Thanks for doing this!

lib/mcp/server.rb
```diff
@@ -313,12 +313,36 @@ def server_info
       }.compact
     end
 
-    def init(request)
+    def init(params)
+      protocol_version = params[:protocolVersion] if params
+      negotiated_version = if protocol_version && Configuration::SUPPORTED_STABLE_PROTOCOL_VERSIONS.include?(protocol_version)
```
koic: On a second review, the check for the presence of `protocol_version` seems to be a redundant condition that can be removed.

```suggestion
      negotiated_version = if Configuration::SUPPORTED_STABLE_PROTOCOL_VERSIONS.include?(protocol_version)
```

lib/mcp/server.rb
```diff
@@ -313,12 +313,36 @@ def server_info
...
+        protocol_version
+      else
+        configuration.protocol_version
+      end
+
+      info = server_info.dup
```
koic: I understand this as processing that occurs during the initial interaction with the server instance, but is it necessary to call `dup`? It looked like `server_info` could be used directly without assigning it to `info`.

duboff: Done, but had to then refactor the rest a bit to not use `delete` since it would mutate the hash

koic:
Thanks! Can you rebase on the latest main branch and resolve the conflicts?

lib/mcp/server.rb
```diff
@@ -313,12 +313,29 @@ def server_info
...
+      protocol_version = params[:protocolVersion] if params
+      negotiated_version = if Configuration::SUPPORTED_STABLE_PROTOCOL_VERSIONS.include?(protocol_version)
+        protocol_version
+      else
+        configuration.protocol_version
+      end
+
+      info = server_info
+      info = info.reject { |key| [:description, :icons].include?(key) } if negotiated_version <= "2025-06-18"
+      info = info.reject { |key| [:title, :websiteUrl].include?(key) } if negotiated_version <= "2025-03-26"
```
koic: The filtering loop logic seems a bit redundant, and the terminology could be clearer. Can you update it?

```diff
diff --git a/lib/mcp/server.rb b/lib/mcp/server.rb
index af3ea15..03628bb 100644
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -18,6 +18,9 @@ module MCP
   class Server
     DEFAULT_VERSION = "0.1.0"
 
+    UNSUPPORTED_PROPERTIES_UNTIL_2025_06_18 = [:description, :icons].freeze
+    UNSUPPORTED_PROPERTIES_UNTIL_2025_03_26 = [:title, :websiteUrl].freeze
+
     class RequestHandlerError < StandardError
       attr_reader :error_type
       attr_reader :original_error
@@ -321,9 +324,10 @@ module MCP
         configuration.protocol_version
       end
 
-      info = server_info
-      info = info.reject { |key| [:description, :icons].include?(key) } if negotiated_version <= "2025-06-18"
-      info = info.reject { |key| [:title, :websiteUrl].include?(key) } if negotiated_version <= "2025-03-26"
+      info = server_info.reject do |property|
+        negotiated_version <= "2025-06-18" && UNSUPPORTED_PROPERTIES_UNTIL_2025_06_18.include?(property) ||
+          negotiated_version <= "2025-03-26" && UNSUPPORTED_PROPERTIES_UNTIL_2025_03_26.include?(property)
+      end
 
       response_instructions = instructions
```

duboff: Done

PR #238, opened 2026-02-24
update base ruby logic and settings for supporting ruby 2.6

shutkovMaksim:
Update code to support ruby 2.6

### DIFF
--- a/mcp.gemspec
+++ b/mcp.gemspec
@@ -31,5 +31,5 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency("json-schema", ">= 4.1")
+  spec.add_dependency("json-schema", ">= 4.1", "< 6.0")
 end


koic:
Why is this support needed?

mcp.gemspec
```diff
@@ -31,5 +31,5 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency("json-schema", ">= 4.1")
+  spec.add_dependency("json-schema", ">= 4.1", "< 6.0")
```
koic, line 34: This would create an undesirable restriction for users on newer Ruby versions, preventing them from using bugfix releases and other updates. Instead of restricting the version, please consider proposing changes to support it on the dependent gem side. https://github.com/voxpupuli/json-schema/pull/561 may be helpful as a reference.

koic:
Support for Ruby 2.6 itself is not opposed, but in that case the dependent gems would also need to support Ruby 2.6. Since the activity around this is unclear, this issue will be closed.

If this is resumed after `json-schema` adds support Ruby 2.6, please reopen it.

PR #246, opened 2026-02-27
Fix `Prompt#validate_arguments!` crash when arguments are `nil`

koic:
## Motivation and Context

`Prompt#validate_arguments!` called `required_args` which does `arguments_value.filter_map`, crashing with `NoMethodError` when `arguments_value` is `nil` (the default for prompts without declared arguments). Additionally, passing `nil` as args caused a crash on `args.keys`.

Fix by defaulting `arguments_value` to `[]` and guarding against `nil` args.

## How Has This Been Tested?

Added a repro test to verify the fix and prevent regression.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/test/mcp/prompt_test.rb
+++ b/test/mcp/prompt_test.rb
@@ -182,7 +182,7 @@ module MCP
       assert_equal expected, FullPrompt.to_h
     end
 
-    test "#to_h handles nil arguments value" do
+    test "#to_h omits arguments key when arguments are not declared" do
       class NoArgumentsPrompt < Prompt
         description "No arguments prompt"
       end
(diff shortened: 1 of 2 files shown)


test/mcp/prompt_test.rb
```diff
@@ -196,6 +196,51 @@ class NoArgumentsPrompt < Prompt
       assert_equal expected, prompt.to_h
     end
 
+    test "#validate_arguments! does not crash when arguments are not declared" do
```
atesgoral: instead of "crash" (which evokes the process crashing and core-dumping) can we say "raise"?

atesgoral (dismissed):

test/mcp/prompt_test.rb
koic: It was updated because that terminology is more appropriate.

PR #252, opened 2026-03-05
[Doc] Fix Rails controller example to return 202 for notifications

### LINKED ISSUE #158: Rails controller example doesn't return 202 on notifications/initialized
**Problem**

When using the [Rails controller example in the README](https://github.com/modelcontextprotocol/ruby-sdk?tab=readme-ov-file#rails-controller) the `notifications/initialized` request returns `200` instead of `202`. 

It doesn't seem like a major thing but when adding the MCP server to Claude it's impossible to add it when the `notifications/initialized` request returns `200` instead of `202`. When updating it to return `202` it works fine.

**Proposed solution**

Getting the status from the transport object instead of `server.handle_json` like this makes it work though, so the updated example would then look like this:

```ruby
class ApplicationController < ActionController::Base
  def index
    server = MCP::Server.new(
      name: "my_server",
      title: "Example Server Display Name", # WARNING: This is a `Draft` and is not supported in the `Version 2025-06-18 (latest)` specification.
      version: "1.0.0",
      instructions: "Use the tools of this server as a last resort",
      tools: [SomeTool, AnotherTool],
      prompts: [MyPrompt],
      server_context: { user_id: current_user.id },
    )
    transport = MCP::Server::Transports::StreamableHTTPTransport.new(server)
    server.transport = transport
    response = transport.handle_request(request)
    status, headers, body = response
    json = JSON.parse(body.first) if body.present? && body.is_a?(Array)
    render(json:, status:, headers:)
  end
end
```

Let me know if you want me to create a PR for it!

koic:
## Motivation and Context

The previous example used `Server#handle_json` directly, which does not control HTTP status codes. Notifications such as `notifications/initialized` require 202 Accepted per the MCP spec, but the example always returned 200.

Use `StreamableHTTPTransport#handle_request` instead, which returns the correct HTTP status code for each request type.

- https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#sending-messages-to-the-server
- https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server

Closes #158

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,11 +242,12 @@ When added to a Rails controller on a route that handles POST requests, your ser
 [Streamable HTTP](https://modelcontextprotocol.io/specification/latest/basic/transports#streamable-http) transport
 requests.
 
-You can use the `Server#handle_json` method to handle requests.
+You can use `StreamableHTTPTransport#handle_request` to handle requests with proper HTTP
+status codes (e.g., 202 Accepted for notifications).
 
 ```ruby
-class ApplicationController < ActionController::Base
-  def index
+class McpController < ActionController::Base
+  def create
     server = MCP::Server.new(
       name: "my_server",
       title: "Example Server Display Name",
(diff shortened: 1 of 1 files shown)


collimarco:
This seems to introduce more boilerplate code... 
Wouldn't be better to fix the original method or introduce a new method?
Syntax before this proposed change was easier to read.

koic:
Yeah, that could be considered. Since this PR is intended to update examples of how the existing API is used, providing a new API would likely be done in a separate PR.

PR #257, opened 2026-03-16
Return 404 for invalid session ID in `handle_regular_request`

koic:
## Motivation and Context

The MCP specification requires that when a server receives a request containing a session ID that is no longer valid, it MUST respond with HTTP 404 Not Found. The `handle_get` path already returned 404 correctly, but `handle_regular_request` (the POST path) was returning 400 "Invalid session ID" instead.

Ref: https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management

> The server MAY terminate the session at any time, after which it MUST respond
> to requests containing that session ID with HTTP 404 Not Found.

## How Has This Been Tested?

Added a test for POST requests with an invalid session ID, verifying that the response is 404 with "Session not found" error message. All existing tests pass.

## Breaking Change

POST requests with an invalid session ID now return HTTP 404 "Session not found" instead of HTTP 400 "Invalid session ID". Clients that match on the 400 status code or the old error message will need to be updated.

However, this change is considered a bug fix because it brings the behavior into compliance with the MCP specification.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/server/transports/streamable_http_transport.rb
+++ b/lib/mcp/server/transports/streamable_http_transport.rb
@@ -261,7 +261,7 @@ module MCP
           unless @stateless
             # If session ID is provided, but not in the sessions hash, return an error
             if session_id && !@sessions.key?(session_id)
-              return [400, { "Content-Type" => "application/json" }, [{ error: "Invalid session ID" }.to_json]]
+              return session_not_found_response
             end
           end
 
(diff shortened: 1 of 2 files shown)


koic:
@modelcontextprotocol/ruby-sdk This may cause merge conflicts with upcoming work, so a review when you have a chance would be appreciated.

PR #268, opened 2026-03-23
Support session expiry controls for `StreamableHTTPTransport`

### LINKED ISSUE #265: Add session timeouts
**Is your feature request related to a problem? Please describe.**

Add session timeouts for [Session ID](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#session-management) (`Mcp-Session-Id`). Session Hijacking [mitigations](https://modelcontextprotocol.io/docs/tutorials/security/security_best_practices#mitigation-4) suggests session expiry to reduce session hijacking issues, this is currently unimplemented in the Ruby SDK.

**Describe the solution you'd like**
Please refer similar implementations in other SDKs:
**go-sdk**
- https://github.com/modelcontextprotocol/go-sdk/blob/main/mcp/streamable.go#L526-L530

**csharp-sdk**
- https://github.com/modelcontextprotocol/csharp-sdk/blob/main/src/ModelContextProtocol.AspNetCore/HttpServerTransportOptions.cs#L110-L138

koic:
## Motivation and Context

The MCP specification recommends expiring session IDs to reduce session hijacking risks: https://modelcontextprotocol.io/specification/latest/basic/security_best_practices#session-hijacking

Several other SDKs (Go, C#, PHP, Python) already implement session expiry controls, but the Ruby SDK did not. Sessions persisted indefinitely unless explicitly deleted or a stream error occurred, leaving abandoned sessions to accumulate in memory.

This adds a `session_idle_timeout:` option to `StreamableHTTPTransport#initialize`. When set, sessions that receive no HTTP requests for the specified duration (in seconds) are automatically expired. Expired sessions return 404 on subsequent requests (GET and POST), matching the MCP specification's behavior for terminated sessions. Each request resets the idle timer, so actively used sessions are not interrupted.

A background reaper thread periodically cleans up expired sessions to handle orphaned sessions that receive no further requests. The reaper only starts when `session_idle_timeout` is configured.

The default is `nil` (no expiry) for backward compatibility, consistent with the Python SDK's approach. The Python SDK recommends 1800 seconds (30 minutes) for most deployments: https://github.com/modelcontextprotocol/python-sdk/pull/2022

Resolves #265.

## How Has This Been Tested?

Added tests for session expiry, idle timeout reset, reaper thread lifecycle, input validation, and default behavior in `streamable_http_transport_test.rb`. All existing tests continue to pass.

## Breaking Change

None. The default value of `session_idle_timeout` is `nil`, which preserves the existing behavior of sessions never expiring. The new `last_active_at` field in the internal session hash is not part of the public API. Existing code that instantiates `StreamableHTTPTransport.new(server)` or
`StreamableHTTPTransport.new(server, stateless: true)` continues to work without changes.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/server/transports/streamable_http_transport.rb
+++ b/lib/mcp/server/transports/streamable_http_transport.rb
@@ -8,18 +8,30 @@ module MCP
   class Server
     module Transports
       class StreamableHTTPTransport < Transport
-        def initialize(server, stateless: false)
+        def initialize(server, stateless: false, session_idle_timeout: nil)
           super(server)
-          # Maps `session_id` to `{ stream: stream_object, server_session: ServerSession }`.
+          # Maps `session_id` to `{ stream: stream_object, server_session: ServerSession, last_active_at: float_from_monotonic_clock }`.
           @sessions = {}
           @mutex = Mutex.new
 
           @stateless = stateless
+          @session_idle_timeout = session_idle_timeout
+
+          if @session_idle_timeout
+            if @stateless
+              raise ArgumentError, "session_idle_timeout is not supported in stateless mode."
+            elsif @session_idle_timeout <= 0
+              raise ArgumentError, "session_idle_timeout must be a positive number."
+            end
+          end
+
+          start_reaper_thread if @session_idle_timeout
         end
 
         REQUIRED_POST_ACCEPT_TYPES = ["application/json", "text/event-stream"].freeze
         REQUIRED_GET_ACCEPT_TYPES = ["text/event-stream"].freeze
         STREAM_WRITE_ERRORS = [IOError, Errno::EPIPE, Errno::ECONNRESET].freeze
+        SESSION_REAP_INTERVAL = 60
 
         def handle_request(request)
           case request.env["REQUEST_METHOD"]
@@ -35,6 +47,9 @@ module MCP
         end
 
         def close
+          @reaper_thread&.kill
+          @reaper_thread = nil
+
           @mutex.synchronize do
             @sessions.each_key { |session_id| cleanup_session_unsafe(session_id) }
           end
@@ -56,6 +71,11 @@ module MCP
               session = @sessions[session_id]
               return false unless session && session[:stream]
 
+              if session_expired?(session)
+                cleanup_session_unsafe(session_id)
+                return false
+              end
+
               begin
                 send_to_stream(session[:stream], notification)
                 true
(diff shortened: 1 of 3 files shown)


lib/mcp/server/transports/streamable_http_transport.rb
```diff
@@ -97,6 +122,36 @@ def send_notification(method, params = nil, session_id: nil)
...
+            @sessions.each_with_object([]) do |(session_id, session), streams|
+              next unless session_expired?(session)
+
+              streams << session[:stream] if session[:stream]
+              @sessions.delete(session_id)
+            end
+          end
+
+          expired_streams.each do |stream|
+            stream.close
```
atesgoral, line 152: Is stream.close idempotent on closed sessions? Can there a be race condition where a session closes by natural means after we collect it for reaping?

koic: `reap_expired_sessions` deletes the session from `@sessions` inside the mutex before calling `stream.close` outside it, so there should be no issue. Added a code comment to clarify this.

lib/mcp/server/transports/streamable_http_transport.rb
```diff
@@ -35,6 +47,9 @@ def handle_request(request)
         end
 
         def close
+          @reaper_thread&.kill
```
atesgoral, line 50: What happens to the mutex if the reaper thread had a lock on it and was busy reaping?

atesgoral (dismissed):
LGTM! Left a question because I don't know how mutexed held by threads behave when the thread is killed while holding a mutex. If this is already safe, no further comments!

lib/mcp/server/transports/streamable_http_transport.rb
koic: As far as I could verify, In MRI Ruby, this should be safe. Ruby runs `ensure` clauses even when a thread is terminated via `Thread#kill`, and `Mutex#synchronize` releases the lock in its `ensure` path.

So if the reaper is killed while holding the mutex, the thread will unwind and the lock should be released as part of that process. As a result, `close`'s `@mutex.synchronize` is not expected to deadlock due to the mutex being left locked by `Thread#kill`.

atesgoral: Thanks! LGTM!

koic:
@atesgoral Thanks for the review. It looks like the approval was dismissed when I resolved a previous conflict. Could you take another look when you have a moment?

PR #231, opened 2026-02-18
Return JSON-RPC protocol errors for unknown tool calls

koic:
## Motivation and Context

Per the MCP specification, an unknown tool call is a protocol-level error, not a tool execution error. Return JSON-RPC -32602 (Invalid params) instead of successful responses with `isError: true`. https://modelcontextprotocol.io/specification/2025-11-25/server/tools#error-handling

Fixes #228.

## How Has This Been Tested?

Added a repro test to verify the fix and prevent regression.

## Breaking Changes

None.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #285, opened 2026-04-01
Fix: return protocol errors for invalid arguments and server errors

### LINKED ISSUE #284: Protocol errors still returned as tool execution errors (invalid arguments, server errors)
**Describe the bug**
Two of the three protocol error cases defined by the spec are still returned as tool execution errors (`isError: true` results) instead of JSON-RPC `error` responses:

1. **Invalid arguments** — missing required arguments and schema validation failures both return `isError: true`
2. **Server errors** — internal exceptions during tool execution return `isError: true`

This is the same class of bug as #228, which was fixed for unknown tool calls in PR #231. The other two cases were not addressed.

**To Reproduce**

1. Define a tool with a required argument
2. Call it with the argument missing

```ruby
server = MCP::Server.new(tools: [MyTool])
response = server.handle({
  jsonrpc: "2.0",
  id: 1,
  method: "tools/call",
  params: { name: "my_tool", arguments: {} }
})
```

3. Response is a successful JSON-RPC result with `isError: true`:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "content": [{"type": "text", "text": "Missing required arguments: required_argument"}],
    "isError": true
  }
}
```

Same behavior for invalid schema validation and internal server errors.

**Expected behavior**

A JSON-RPC protocol error, consistent with how unknown tool calls are handled:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32602,
    "message": "Invalid params",
    "data": "Missing required arguments: required_argument"
  }
}
```

**Additional context**

The [MCP spec](https://modelcontextprotocol.io/specification/2024-11-05/server/tools#error-handling) draws a clear line between protocol errors and tool execution errors:

> **Protocol Errors:** Standard JSON-RPC errors for issues like:
> - Unknown tools
> - Invalid arguments
> - Server errors

> **Tool Execution Errors:** Reported in tool results with `isError: true`:
> - API failures
> - Invalid input data
> - Business logic errors

**A note on server errors:** #159 raised a valid concern that tool execution errors like `ActiveRecord::RecordNotFound` shouldn't be protocol errors. But the fix for that is for the tool to catch those exceptions and return `Tool::Response.new(content, error: true)` — that's the tool reporting a known failure. If an exception bubbles up uncaught, the server can't distinguish a business error from an actual crash, so treating it as a server error (protocol error) is the correct default.

The trail:

1. PR #165 moved all three protocol error cases from `RequestHandlerError` (JSON-RPC errors) to `error_tool_response` (`isError: true`)
2. #228 flagged this as a spec violation
3. PR #231 fixed unknown tools only — invalid arguments and server errors still return `isError: true`

The remaining two cases in `Server#call_tool` need to raise `RequestHandlerError` instead of returning `error_tool_response`, matching the pattern PR #231 restored for unknown tools.

sqrlington:
## Motivation and Context

PR #231 fixed unknown tool calls to return protocol errors per the MCP spec, but two of the three protocol error cases still return `isError: true` instead of JSON-RPC errors:

- **Invalid arguments** — missing required arguments and schema validation failures return `isError: true` instead of `-32602`
- **Server errors** — uncaught exceptions during tool execution return `isError: true` instead of `-32603`

Per the [MCP spec](https://modelcontextprotocol.io/specification/2024-11-05/server/tools#error-handling), these are protocol errors, not tool execution errors.

Fixes #284.

## How Has This Been Tested?

Updated existing tests to assert JSON-RPC error responses instead of `isError: true` results. All tests pass, rubocop clean.

## Breaking Changes

Clients relying on `isError: true` for invalid arguments or server errors will now receive JSON-RPC error responses instead. This matches the behavior PR #231 introduced for unknown tools.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -316,13 +316,12 @@ module MCP
           add_instrumentation_data(client: client) if client
 
           result
+        rescue RequestHandlerError => e
+          report_exception(e.original_error || e, { request: request })
+          add_instrumentation_data(error: e.error_type)
+          raise e
         rescue => e
           report_exception(e, { request: request })
-          if e.is_a?(RequestHandlerError)
-            add_instrumentation_data(error: e.error_type)
-            raise e
-          end
-
           add_instrumentation_data(error: :internal_error)
           raise RequestHandlerError.new("Internal error handling #{method} request", request, original_error: e)
         end
@@ -421,7 +420,7 @@ module MCP
         add_instrumentation_data(error: :missing_required_arguments)
 
         missing = tool.input_schema.missing_required_arguments(arguments).join(", ")
-        return error_tool_response("Missing required arguments: #{missing}")
+        raise RequestHandlerError.new("Missing required arguments: #{missing}", request, error_type: :invalid_params)
       end
 
       if configuration.validate_tool_call_arguments && tool.input_schema
@@ -430,7 +429,7 @@ module MCP
         rescue Tool::InputSchema::ValidationError => e
           add_instrumentation_data(error: :invalid_schema)
 
-          return error_tool_response(e.message)
+          raise RequestHandlerError.new(e.message, request, error_type: :invalid_params)
         end
       end
 
@@ -440,9 +439,12 @@ module MCP
     rescue RequestHandlerError
       raise
     rescue => e
-      report_exception(e, request: request)
-
-      error_tool_response("Internal error calling tool #{tool_name}: #{e.message}")
+      raise RequestHandlerError.new(
+        "Internal error calling tool #{tool_name}: #{e.message}",
+        request,
+        error_type: :internal_error,
+        original_error: e,
+      )
     end
 
     def list_prompts(request)
--- a/test/mcp/server_test.rb
+++ b/test/mcp/server_test.rb
@@ -312,7 +312,7 @@ module MCP
       assert_instrumentation_data({ method: "tools/call", tool_name: tool_name, tool_arguments: tool_args })
     end
 
-    test "#handle tools/call returns error response with isError true if required tool arguments are missing" do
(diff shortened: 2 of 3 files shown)


lib/mcp/server.rb
koic, line 324: `call_tool` no longer calls `report_exception` before re-raising. The rescue here now receives `RequestHandlerError` instead of the original exception, so `exception_reporter` loses the original
   class and backtrace.

```diff
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -317,7 +317,8 @@ module MCP

           result
         rescue => e
-          report_exception(e, { request: request })
+          reported = e.is_a?(RequestHandlerError) ? (e.original_error || e) : e
+          report_exception(reported, { request: request })
           if e.is_a?(RequestHandlerError)
             add_instrumentation_data(error: e.error_type)
             raise e
```

test/mcp/server_test.rb
```diff
@@ -428,13 +418,18 @@ def call(message:, server_context: nil)
         id: 1,
       }
 
+      @server.configuration.exception_reporter.expects(:call).with do |exception, server_context|
+        assert_not_nil exception
```
koic: With the unwrap fix above, this should verify that `exception_reporter` receives the original exception, not the `RequestHandlerError` wrapper.

```suggestion
        refute_kind_of MCP::Server::RequestHandlerError, exception
```

lib/mcp/server.rb
sqrlington: Took the liberty of refactoring the rescue block to avoid the duplicated `is_a?` checks when I implemented this suggestion

koic: Ah, great! Thanks!

PR #290, opened 2026-04-03
[Doc] Clarify that `StreamableHTTPTransport` requires a single-process server

koic:
## Motivation and Context

`StreamableHTTPTransport` stores session and SSE stream state in memory using instance variables (`@sessions`, `@mutex`). This design requires all HTTP requests to be handled within the same process to share state.

Process-based servers like Unicorn fork separate worker processes that do not share memory, which breaks session management and SSE connections. This requirement is not obvious from the current documentation.

## How Has This Been Tested?

Documentation-only change.

## Breaking Changes

None.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,13 @@ transport = MCP::Server::Transports::StreamableHTTPTransport.new(server, session
 
 ### Usage
 
+> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


atesgoral:
We could probably introduce a session store adapter later

PR #298, opened 2026-04-05
[Doc] Reorganize README to surface key information earlier

koic:
## Motivation and Context

While working on #295 and #297, it became clear that the current README.md does not surface the information users are likely looking for at the top. As a result, I have reorganized the overall structure from a user’s perspective.

There is likely still room for further improvements, but even at this stage, the structure should be more useful than before.

This PR intentionally splits the changes into three commits.

1. Custom methods are an advanced feature and may not be appropriate for the top of the README.md. Move them to the bottom of the page.
2. Users are more likely to want to see Usage first, rather than a feature introduction like Sampling. Sampling has been moved below Resource Templates so that Usage appears at the top.
3. For Usage, stdio is easier to try first than a Rails controller. Moved stdio to the top and placed Rails Controller after it.

Only sections were reorganized with headings. No content was changed.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,43 +56,6 @@ It implements the Model Context Protocol specification, handling model context r
 
 ### Usage
 
-> [!IMPORTANT]
-> `MCP::Server::Transports::StreamableHTTPTransport` stores session and SSE stream state in memory,
-> so it must run in a single process. Use a single-process server (e.g., Puma with `workers 0`).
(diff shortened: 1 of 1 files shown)


koic:
It would be easier to resolve conflicts if this is merged before #295, #297, and #299.

PR #309, opened 2026-04-11
Add `around_request` hook for request instrumentation

### LINKED ISSUE #302: Proposal: Instrumentation Middleware
I'm proposing that we implement optional instrumentation middleware hooks for tool calls within the MCP Server. 

Some context: We are accumulating a health list of tools available in platform, but we currently cannot create spans for the tool calls in our telemetry vendor (Datadog). We would like to be able to wrap the block of tool calls in...

```ruby
Datadog::Tracing.trace(name, tags:) {  }
```

I believe this should either live in `lib/mcp/instrumentation.rb` (instrumenting all `handle_request` calls) or more directly in `lib/mcp/server.rb` within the `#call_tool` method.

If this is interesting and aligned with the goals of this gem, we can take a pass at the initial PR here. Thanks!

cc @paholg

koic:
## Motivation and Context

`instrumentation_callback` fires after request execution, making it impossible to wrap requests with Application Performance Monitoring (APM) spans (e.g. Datadog tracing). This adds `around_request`, which wraps request handling and allows executing code before and after each request.

For example, to wrap requests with a Datadog trace:

```ruby
MCP.configure do |config|
  config.around_request = ->(data, &request_handler) {
    Datadog::Tracing.trace("mcp.#{data[:method]}") do
      request_handler.call
    end
  }
end
```

`instrumentation_callback` is soft-deprecated in favor of `around_request`.

## How Has This Been Tested?

Added tests for `around_request` in configuration, instrumentation, and server test files. All existing tests continue to pass.

## Breaking Changes

`around_request` itself is optional and defaults to a pass-through, and `instrumentation_callback` continues to work as before.

However, `instrument_call` now accepts a `server_context:` keyword and its block is invoked without arguments (previously `yield block` passed the block Proc as an argument). This affects anyone mixing `MCP::Instrumentation` into their own classes, though the public `Server` API is unchanged.

Resolves #302.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,15 +159,17 @@ MCP.configure do |config|
     end
   }
 
-  config.instrumentation_callback = ->(data) {
-    puts "Got instrumentation data #{data.inspect}"
+  config.around_request = ->(data, &request_handler) {
+    logger.info("Start: #{data[:method]}")
+    request_handler.call
+    logger.info("Done: #{data[:method]}, tool: #{data[:tool_name]}")
   }
 end
 ```
 
 or by creating an explicit configuration and passing it into the server.
 This is useful for systems where an application hosts more than one MCP server but
-they might require different instrumentation callbacks.
+they might require different configurations.
 
 ```ruby
 configuration = MCP::Configuration.new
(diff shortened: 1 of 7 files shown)


atesgoral (approved):
Nice!

PR #321, opened 2026-04-18
Add MCP Streamable HTTP specification support for the client

atesgoral:
## Summary

Implements the [MCP Streamable HTTP specification](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#streamable-http) for the Ruby SDK client. Rebased from #210 (original work by @keisku) onto current `main`, with merge conflict resolution, bug fixes, and all review feedback addressed.

**Original PR:** #210 by @keisku — thank you for kicking off this effort!

**Already supported (before this PR):**
- POST with JSON body
- `Accept: application/json, text/event-stream` header
- Parse `application/json` response

**This PR adds:**
- Parse `text/event-stream` (SSE) response via `event_stream_parser` gem
- Session management (`MCP-Session-Id` header)
- Handle 404 as session expiration
- Handle 202 Accepted when server delivers response via SSE stream
- DELETE for session termination
- Protocol version header (`MCP-Protocol-Version`)
- Client conformance test runner (`conformance/client.rb`)

## Changes from #210

- Rebased onto current `main` (resolved merge conflicts with faraday_customizer and test changes)
- Made `event_stream_parser` an optional dependency (like Faraday), per @koic's [feedback](https://github.com/modelcontextprotocol/ruby-sdk/pull/210#discussion_r2677983134) — users add it to their Gemfile when needed
- Addressed all of @koic's review comments (TODO descriptions, consistent header naming, gemspec vs Gemfile dependency placement)
- Fixed 202 Accepted response handling when server delivers via SSE stream

## Commits

1. **Add MCP Streamable HTTP specification support for the client** — @keisku's original work, squashed and rebased
2. **Handle 202 Accepted response when server delivers via SSE stream** — bug fix
3. **Address PR review feedback from Koichi** — review feedback
4. **Make event_stream_parser an optional dependency like Faraday** — addresses @koic's latest feedback

## Test plan

- [x] `ruby -I lib -I test test/mcp/client_test.rb` — 40 runs, 125 assertions, 0 failures
- [x] `ruby -I lib -I test test/mcp/client/http_test.rb` — 20 runs, 48 assertions, 0 failures
- [x] Full test suite: 704 runs, 1762 assertions, 0 failures
- [x] RuboCop clean on changed files

## Pending TODOs (out of scope)

- [Client GET for SSE stream](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#streamable-http)
[...]

PR #322, opened 2026-04-18
Parse SSE responses in HTTP client via event_stream_parser

atesgoral:
## Summary

First slice of #321 (Streamable HTTP client support), factored out for easier review.

The [Streamable HTTP spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#streamable-http) allows servers to respond to a POST with either `application/json` or `text/event-stream`. `MCP::Client::HTTP` currently rejects the latter. This PR adds SSE parsing via the optional [`event_stream_parser`](https://rubygems.org/gems/event_stream_parser) gem — users add it to their Gemfile only when their server uses SSE, matching the existing pattern for `faraday`.

`send_request` now dispatches on the response `Content-Type`, scans the event stream for the first JSON-RPC response message (skipping comments, notifications, and non-response events), and returns it.

This PR intentionally does **not** include other Streamable HTTP features — those will land as follow-up stacked PRs:

- 202 Accepted handling for async-delivered responses
- `Mcp-Session-Id` header tracking / `SessionExpiredError` on 404
- `Client#connect` handshake helper + `protocol_version` tracking
- `Client#close` / DELETE session termination
- Example + docs rewrite

Splitting it this way keeps each PR self-contained and shippable: after this merges, the client correctly handles both JSON and SSE responses from stateless servers.

## Changes

- `lib/mcp/client/http.rb`: replace `validate_response_content_type!` with a `parse_response_body` dispatch on `Content-Type`; add `parse_sse_response` and `require_event_stream_parser!`
- `Gemfile`: add `event_stream_parser` to the test group
- `docs/building-clients.md`: mention `event_stream_parser` as an optional dependency for SSE responses
- `test/mcp/client/http_test.rb`: add tests for SSE happy path, SSE error response, SSE stream with no response message, and LoadError when the gem is missing; convert the old "non-JSON response" test to use `text/html` (since `text/event-stream` is now valid)

## Test plan

- [x] `rake test` passes (737 runs, 1847 assertions)
- [x] `rake rubocop` shows no new offenses on changed files
- [x] New tests cover SSE happy path, SSE error responses, empty SSE streams, and missing `event_stream_parser` gem

### DIFF
--- a/docs/building-clients.md
+++ b/docs/building-clients.md
@@ -51,11 +51,12 @@ stdio_transport.close
 
 ## HTTP Transport
 
-Use `MCP::Client::HTTP` to interact with MCP servers over HTTP. Requires the `faraday` gem:
+Use `MCP::Client::HTTP` to interact with MCP servers over HTTP. Requires the `faraday` gem, plus `event_stream_parser` if the server uses SSE (`text/event-stream`) responses:
 
 ```ruby
 gem 'mcp'
 gem 'faraday', '>= 2.0'
+gem 'event_stream_parser', '>= 1.0' # optional, required only for SSE responses
 ```
 
 ```ruby
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,7 @@ gem "yard", "~> 0.9"
 gem "yard-sorbet", "~> 0.9" if RUBY_VERSION >= "3.1"
 
 group :test do
+  gem "event_stream_parser", ">= 1.0"
   gem "faraday", ">= 2.0"
   gem "minitest", "~> 5.1", require: false
   gem "mocha"
--- a/README.md
+++ b/README.md
@@ -1599,11 +1599,12 @@ The stdio transport automatically handles:
 
 Use the `MCP::Client::HTTP` transport to interact with MCP servers using simple HTTP requests.
 
-You'll need to add `faraday` as a dependency in order to use the HTTP transport layer:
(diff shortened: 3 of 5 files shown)


docs/building-clients.md
```diff
@@ -51,11 +51,12 @@ stdio_transport.close
 
 ## HTTP Transport
 
-Use `MCP::Client::HTTP` to interact with MCP servers over HTTP. Requires the `faraday` gem:
+Use `MCP::Client::HTTP` to interact with MCP servers over HTTP. Requires the `faraday` gem, plus `event_stream_parser` if the server uses SSE (`text/event-stream`) responses:
 
 ```ruby
 gem 'mcp'
 gem 'faraday', '>= 2.0'
+gem 'event_stream_parser', '>= 1.0' # optional, required only for SSE responses
```
koic, line 59: It would be helpful to have similar documentation in the README.md as well.
https://github.com/modelcontextprotocol/ruby-sdk#http-transport-layer

atesgoral: Good call — added the same note to the README's HTTP Transport Layer section in the new commit.

koic:
Looks good to me. Can you squash commits into one?

atesgoral:
Squashed to a single commit as requested. E2E tested as well.

koic (approved):

koic:
Thanks!

PR #323, opened 2026-04-18
Handle 202 Accepted response in HTTP client

atesgoral:
## Summary

Second slice of #321 (Streamable HTTP client support).

The [Streamable HTTP spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server) allows a server to respond `202 Accepted` when it has received the request but will deliver the response later via an SSE stream. The client currently errors on 202 because the response has no parseable `Content-Type`. Return `{ "accepted" => true }` so callers can proceed rather than hard-fail on a valid server response.

Actually picking up the deferred response requires listening on an SSE stream (GET-for-SSE), which is a TODO in the main PR and out of scope here. This change is just the graceful-no-hard-error part.

## Changes

- `lib/mcp/client/http.rb`: add `elsif response.status == 202` branch to `parse_response_body` returning `{ "accepted" => true }`
- `test/mcp/client/http_test.rb`: test for 202 response with empty body

## Test plan

- [x] `rake test` passes (738 runs, 1848 assertions)
- [x] `rake rubocop` no new offenses on changed files
- [x] New test covers 202 response with empty body

## Stack

- #322 — SSE parsing via event_stream_parser
- **This PR** — 202 Accepted handling
- (next) — stateful client: `connect`, session-ID tracking, `close`/DELETE

### DIFF
--- a/lib/mcp/client/http.rb
+++ b/lib/mcp/client/http.rb
@@ -100,6 +100,10 @@ module MCP
       end
 
       def parse_response_body(response, method, params)
+        # 202 Accepted is the server's ACK for a JSON-RPC notification or response; no body is expected.
+        # https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server
+        return if response.status == 202
+
         content_type = response.headers["Content-Type"]
 
         if content_type&.include?("text/event-stream")
--- a/examples/streamable_http_server.rb
+++ b/examples/streamable_http_server.rb
@@ -171,7 +171,7 @@ puts <<~MESSAGE
      curl -i http://localhost:9393 -H "Mcp-Session-Id: YOUR_SESSION_ID" \\
        --json '{"jsonrpc":"2.0","method":"tools/call","id":3,"params":{"name":"notification_tool","arguments":{"message":"Hello SSE!", "delay": 2}}}'
 
-  Note: When an SSE stream is active, tool responses will appear in the SSE stream and the POST request will return {"accepted": true}
+  Note: When an SSE stream is active, tool responses will appear in the SSE stream and the POST request will return 202 Accepted with no body.
 
   Press Ctrl+C to stop the server
 MESSAGE
--- a/test/mcp/client/http_test.rb
+++ b/test/mcp/client/http_test.rb
@@ -369,6 +369,21 @@ module MCP
         assert_equal("Invalid request", response.dig("error", "message"))
       end
 
+      def test_send_request_returns_nil_for_202_accepted_response
+        request = {
+          jsonrpc: "2.0",
+          method: "notifications/initialized",
+        }
+
+        stub_request(:post, url)
+          .with(body: request.to_json)
+          .to_return(status: 202, body: "")
+
+        response = client.send_request(request: request)
+
+        assert_nil(response)
+      end
+
       def test_send_request_raises_error_for_sse_without_response
         request = {
           jsonrpc: "2.0",


docs/building-clients.md
```diff
@@ -51,11 +51,12 @@ stdio_transport.close
 
 ## HTTP Transport
 
-Use `MCP::Client::HTTP` to interact with MCP servers over HTTP. Requires the `faraday` gem:
+Use `MCP::Client::HTTP` to interact with MCP servers over HTTP. Requires the `faraday` gem, plus `event_stream_parser` if the server uses SSE (`text/event-stream`) responses:
 
 ```ruby
 gem 'mcp'
 gem 'faraday', '>= 2.0'
+gem 'event_stream_parser', '>= 1.0' # optional, required only for SSE responses
```
koic: It would be helpful to have similar documentation in the README.md as well.
https://github.com/modelcontextprotocol/ruby-sdk#http-transport-layer

lib/mcp/client/http.rb
```diff
@@ -106,6 +106,10 @@ def parse_response_body(response, method, params)
           parse_sse_response(response.body, method, params)
         elsif content_type&.include?("application/json")
           response.body
+        elsif response.status == 202
+          # Server accepted the request and will deliver the response via an SSE stream.
+          # https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server
+          { "accepted" => true }
```
koic: I think the [spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server) defines 202 as the ACK for a `notification/response`, rather than "request accepted, SSE coming later". The [TypeScript](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/streamableHttp.ts) and [Python](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/streamable_http.py) SDKs also check `status == 202` before Content-Type and return void rather than a Hash sentinel. In the current diff, the 202 branch comes last, so a 202 with any Content-Type header would skip it; the existing test happens to pass only because WebMock leaves Content-Type unset. Also, `{ "accepted" => true }` overlaps with the Hash shape of real JSON-RPC responses. What do you think about something closer to this?

```diff
 def parse_response_body(response, method, params)
+  # 202 Accepted is the ACK for a notification/response; no body is expected.
+  # https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server
+  return if response.status == 202
+
   content_type = response.headers["Content-Type"]

   if content_type&.include?("text/event-stream")
     parse_sse_response(response.body, method, params)
   elsif content_type&.include?("application/json")
     response.body
-  elsif response.status == 202
-    # Server accepted the request and will deliver the response via an SSE stream.
-    # https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server
-    { "accepted" => true }
   else
     raise RequestHandlerError.new(...)
   end
 end
```

The tests will need a matching update as well.

atesgoral: Verified against the spec and the TS/Python SDKs. Updated the diff to check `status == 202` before Content-Type dispatch and return `nil` instead of the Hash sentinel. Also corrected the commit message (202 is the ACK for a notification/response, not "SSE coming later") and cleaned up a stale doc string in `examples/streamable_http_server.rb`. Thanks for catching this!

koic (approved):

koic:
Thanks!

PR #325, opened 2026-04-20
Track Mcp-Session-Id and protocol version in HTTP client

atesgoral:
## Summary

- Capture `Mcp-Session-Id` and `protocolVersion` from the `initialize` response and expose them on the transport as `session_id` / `protocol_version`
- Automatically include `Mcp-Session-Id` on subsequent requests once captured
- Map 404 responses to a new `MCP::Client::SessionExpiredError` (subclass of `RequestHandlerError`) and clear local session state so callers can start a new session

Per the Streamable HTTP spec:

> If the server returns an `Mcp-Session-Id` header during initialization, clients MUST include it on all subsequent HTTP requests.

> When a client receives HTTP 404 in response to a request containing the `Mcp-Session-Id`, it MUST start a new session by sending a new `InitializeRequest` without a session ID attached.

https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management

This is part of a series of PRs splitting up #321. Previous: #322 (SSE parsing) and #323 (202 Accepted). Next: session termination via DELETE (`close`), then client handshake ergonomics and example rewrite.

## Test plan

- [x] New tests for session capture on `initialize`
- [x] New tests that session header is included on subsequent requests
- [x] New test that session ID is not overwritten by later responses
- [x] New test that stateless servers (no `Mcp-Session-Id` header) work fine
- [x] 404 now raises `SessionExpiredError` and clears session state
- [x] Full suite passes (`bundle exec rake test`)

### DIFF
--- a/lib/mcp/client/http.rb
+++ b/lib/mcp/client/http.rb
@@ -1,24 +1,42 @@
 # frozen_string_literal: true
 
+require_relative "../methods"
+
 module MCP
   class Client
+    # TODO: HTTP GET for SSE streaming is not yet implemented.
+    #   https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#listening-for-messages-from-the-server
+    # TODO: Resumability and redelivery with Last-Event-ID is not yet implemented.
+    #   https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#resumability-and-redelivery
     class HTTP
       ACCEPT_HEADER = "application/json, text/event-stream"
+      SESSION_ID_HEADER = "Mcp-Session-Id"
+      PROTOCOL_VERSION_HEADER = "MCP-Protocol-Version"
 
-      attr_reader :url
+      attr_reader :url, :session_id, :protocol_version
 
       def initialize(url:, headers: {}, &block)
         @url = url
         @headers = headers
         @faraday_customizer = block
+        @session_id = nil
+        @protocol_version = nil
       end
 
+      # Sends a JSON-RPC request and returns the parsed response body.
+      # After a successful `initialize` handshake, the session ID and protocol
+      # version returned by the server are captured and automatically included
+      # on subsequent requests.
       def send_request(request:)
         method = request[:method] || request["method"]
         params = request[:params] || request["params"]
 
-        response = client.post("", request)
-        parse_response_body(response, method, params)
+        response = client.post("", request, session_headers)
+        body = parse_response_body(response, method, params)
+
+        capture_session_info(method, response, body)
+
+        body
       rescue Faraday::BadRequestError => e
         raise RequestHandlerError.new(
           "The #{method} request is invalid",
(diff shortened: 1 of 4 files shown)


lib/mcp/client/http.rb
```diff
@@ -41,10 +59,13 @@ def send_request(request:)
           original_error: e,
         )
       rescue Faraday::ResourceNotFound => e
-        raise RequestHandlerError.new(
+        # Per spec, the server MAY terminate the session at any time and MUST
+        # respond with 404 to requests containing an expired session ID.
+        # https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management
+        clear_session
+        raise SessionExpiredError.new(
```
koic: This implementation currently raises `SessionExpiredError` for every 404, including when `@session_id` is `nil` (e.g., the very first request before `initialize`, or a stateless server where the URL is misconfigured).

The spec wording is:

> When a client receives HTTP 404 in response to a request **containing an `Mcp-Session-Id`**, it MUST start a new session...

So strictly speaking, "session expired" would apply only when a session ID was actually attached.

Not a blocker, and it's fine to keep the current behavior if that's intentional, but one alternative would be:

```ruby
if @session_id
  clear_session
  raise SessionExpiredError.new(...)
else
  raise RequestHandlerError.new(..., error_type: :not_found, ...)
end
```

That way a 404 from a wrong URL or a server that's simply not there surfaces as a generic `:not_found` rather than as a potentially misleading "session expired".

Would this make sense?

lib/mcp/client/http.rb
```diff
@@ -1,24 +1,42 @@
...
+
 module MCP
   class Client
+    # TODO: HTTP GET for SSE streaming is not yet implemented.
+    #   https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#listening-for-messages-from-the-server
+    # TODO: Resumability and redelivery with Last-Event-ID is not yet implemented.
+    #   https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#resumability-and-redelivery
     class HTTP
       ACCEPT_HEADER = "application/json, text/event-stream"
+      SESSION_ID_HEADER = "Mcp-Session-Id"
```
atesgoral, line 13: The spec has this as "MCP-Session-Id". Will standardize casing for existing server code + this new client code in a separate PR.

lib/mcp/client/http.rb
atesgoral: Good catch, the spec scopes the "start a new session" MUST to 404s carrying an `MCP-Session-Id`, so the session-expired interpretation shouldn't apply to plain 404s. Updated per your suggestion: branch on `@session_id` and only raise `SessionExpiredError` (and clear session state) when a session was attached; otherwise raise a generic `RequestHandlerError` with `:not_found`. Split the test into pre-session and post-session cases.

lib/mcp/client/http.rb
koic: Good catch, it looks like this was updated in the spec at some point.
https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1325

koic (approved):

koic:
Thanks!

PR #326, opened 2026-04-20
Add HTTP client close for explicit session termination

atesgoral:
## Summary

Adds `MCP::Client::HTTP#close` to explicitly terminate a server session.

Per the [Streamable HTTP spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management), clients that no longer need a session SHOULD send an HTTP DELETE to the MCP endpoint with the `Mcp-Session-Id` header. The server MAY respond with 405 Method Not Allowed when it doesn't support client-initiated termination.

## Behavior

- Sends `DELETE` to the MCP endpoint with `Mcp-Session-Id` and `MCP-Protocol-Version` headers
- Clears local session state (`session_id`, `protocol_version`) regardless of the server's response
- `405 Method Not Allowed` is treated as success
- Other Faraday errors are swallowed so local cleanup always succeeds
- No-op when no session has been established

## Part of a stacked series

This is PR #4 of a series splitting #321 into small, self-contained changes:

- [x] #322 SSE parsing
- [x] #323 `202 Accepted` handling
- [x] #325 Session tracking (`Mcp-Session-Id` / `MCP-Protocol-Version`)
- [x] **this PR — Session termination via DELETE**
- [ ] Client handshake ergonomics (`connect`)
- [ ] Example rewrite + expanded docs

## Test plan

- [x] New unit tests cover: DELETE sent with session headers, local state cleared, no-op without session, tolerance of 405 and 5xx, idempotency
- [x] Full test suite passes (`bundle exec rake test` — 762 runs)
- [x] E2E verified against `examples/streamable_http_server.rb`: DELETE 200 from server, second request with the terminated session ID raises `SessionExpiredError`

### DIFF
--- a/docs/building-clients.md
+++ b/docs/building-clients.md
@@ -85,6 +85,12 @@ http_transport.protocol_version # => "2025-11-25"
 
 If the server terminates the session, subsequent requests return HTTP 404 and the transport raises `MCP::Client::SessionExpiredError` (a subclass of `RequestHandlerError`). Session state is cleared automatically; callers should start a new session by sending a fresh `initialize` request.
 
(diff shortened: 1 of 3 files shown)


koic (approved):

koic:
Thanks!

PR #327, opened 2026-04-20
Add client-level connect for initialize handshake

atesgoral:
## Summary

Adds `MCP::Client#connect` to perform the MCP [initialization handshake](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization): sends an `initialize` request through the transport, followed by the required `notifications/initialized` notification, and returns the server's `InitializeResult` (protocol version, capabilities, server info, instructions).

- `client.connect(client_info:, protocol_version:, capabilities:)` — idempotent. Delegates to `transport.connect(...)` when the transport exposes an explicit handshake (e.g. `MCP::Client::HTTP`); no-op when it doesn't (e.g. `MCP::Client::Stdio`, which still initializes implicitly on the first `send_request`).
- `client.connected?` — reports handshake completion; delegates to the transport when supported, otherwise returns `true`.
- `client.server_info` — returns the transport's cached `InitializeResult`, or `nil` when the transport doesn't expose it.

### Why on the client

`initialize` is a protocol method — same level as `ping`, `list_tools`, `call_tool` — and the client already owns those. Python (`ClientSession.initialize()`) and TypeScript (`Client.connect(transport)`) both put the handshake on the client. Aligning Ruby with those SDKs gives users a unified API regardless of transport and keeps the transport layer focused on byte transport plus transport-specific state (e.g. HTTP session headers).

HTTP transport retains `connect`, `connected?`, `server_info`, `session_id`, and `protocol_version` — the client delegates through. The existing `capture_session_info` hook still fires during the handshake because `initialize` goes through `transport.send_request`, so session headers and negotiated protocol version continue to get captured automatically.

### Stdio

Stdio currently initializes implicitly on the first `send_request` via a private `initialize_session` — unchanged in this PR. `client.connect` is a no-op for stdio, and existing stdio clients continue working without modification.

**Open question for SDK contributors**: should we fully unify by removing stdio's implicit init and requiring `client.connect` universally? It would tighten the contract (one explicit handshake path) at the cost of a breaking change for existing stdio users. Happy to do that in a follow-up if there's consensus.

### Example

```ruby
transport = MCP::Client::HTTP.new(url: "https://api.example.com/mcp")
client = MCP::Client.new(transport: transport)
client.connect
# => { "protocolVersion" => "2025-11-25", "capabilities" => {...}, "serverInfo" => {...} }

client.connected?  # => true
client.server_info # => cached InitializeResult
client.tools
transport.close
```

## Test plan

- [x] Client-level unit tests: delegation, keyword-arg forwarding, no-op when transport lacks `:connect`, `connected?` delegation and default, `server_info` delegation and default
- [x] Transport-level unit tests (from the previous revision of this PR): happy path, defaults, custom parameters, idempotence, JSON-RPC error response, missing `result`, `connected?` lifecycle, reconnect after `close`
- [x] `rake test` — 782 runs, 1938 assertions, 0 failures
- [x] `rubocop` clean on touched files
- [x] E2E against `examples/streamable_http_server.rb`: `client.connect` → tools → call_tool → `transport.close` → reconnect via `client.connect` (new session id issued); `client.server_info` correctly clears on `close`
- [x] `examples/stdio_client.rb` still runs end-to-end without an explicit `client.connect` call (implicit-init path unchanged)

### DIFF
--- a/lib/mcp/client/http.rb
+++ b/lib/mcp/client/http.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
+require "securerandom"
+require_relative "../../json_rpc_handler"
+require_relative "../configuration"
 require_relative "../methods"
+require_relative "../version"
 
 module MCP
   class Client
@@ -13,7 +17,7 @@ module MCP
       SESSION_ID_HEADER = "Mcp-Session-Id"
       PROTOCOL_VERSION_HEADER = "MCP-Protocol-Version"
 
-      attr_reader :url, :session_id, :protocol_version
+      attr_reader :url, :session_id, :protocol_version, :server_info
 
       def initialize(url:, headers: {}, &block)
         @url = url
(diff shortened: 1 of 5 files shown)


lib/mcp/client/http.rb
```diff
@@ -13,14 +17,86 @@ class HTTP
...
+
+        unless response.is_a?(Hash) && response["result"].is_a?(Hash)
+          raise RequestHandlerError.new(
+            "Server initialization failed: missing result in response",
+            { method: MCP::Methods::INITIALIZE },
+            error_type: :internal_error,
+          )
+        end
+
+        @server_info = response["result"]
```
koic, line 86: According to the MCP specification, it seems advisable to validate the negotiated protocol version. The following is an example.

```diff
+
+        negotiated_protocol_version = @server_info["protocolVersion"]
+        unless MCP::Configuration::SUPPORTED_STABLE_PROTOCOL_VERSIONS.include?(negotiated_protocol_version)
+          # Per spec, if the client does not support the server's returned protocol version,
+          # it SHOULD disconnect. Roll back state captured on the initialize response before raising,
+          # so a retry starts from a clean slate.
+          # https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#version-negotiation
+          clear_session
+          raise RequestHandlerError.new(
+            "Server initialization failed: unsupported protocol version #{negotiated_protocol_version.inspect}",
+            { method: MCP::Methods::INITIALIZE },
+            error_type: :internal_error,
+          )
+        end
```

lib/mcp/client/http.rb
```diff
@@ -13,14 +17,86 @@ class HTTP
...
+          id: SecureRandom.uuid,
+          method: MCP::Methods::INITIALIZE,
+          params: {
+            protocolVersion: protocol_version,
+            capabilities: capabilities,
+            clientInfo: client_info,
+          },
+        })
+
+        if response.is_a?(Hash) && response.key?("error")
```
koic, line 67: It seems reasonable to clear the session when an error occurs. The following is an example.

```diff
         if response.is_a?(Hash) && response.key?("error")
+          # `send_request` has already invoked `capture_session_info`, which may have
+          # stored a session id from the response headers even though the server then
+          # returned a JSON-RPC error. Clear it so a retry starts fresh.
+          clear_session
           error = response["error"]
           raise RequestHandlerError.new(
             "Server initialization failed: #{error["message"]}",
@@ -74,6 +78,7 @@ module MCP
         end

         unless response.is_a?(Hash) && response["result"].is_a?(Hash)
+          clear_session
           raise RequestHandlerError.new(
             "Server initialization failed: missing result in response",
             { method: MCP::Methods::INITIALIZE },
@@ -97,10 +102,15 @@ module MCP
           )
         end

-        send_request(request: {
-          jsonrpc: JsonRpcHandler::Version::V2_0,
-          method: MCP::Methods::NOTIFICATIONS_INITIALIZED,
-        })
+        begin
+          send_request(request: {
+            jsonrpc: JsonRpcHandler::Version::V2_0,
+            method: MCP::Methods::NOTIFICATIONS_INITIALIZED,
+          })
+        rescue
+          clear_session
+          raise
+        end

         @connected = true
         @server_info
```

koic:
> **Open question for SDK contributors**: should we fully unify by removing stdio's implicit init and requiring `client.connect` universally? It would tighten the contract (one explicit handshake path) at the cost of a breaking change for existing stdio users.

It seems reasonable to provide `client.connect` for stdio as well. To minimize user impact while avoiding a breaking change, proceeding in the following steps looks appropriate:

1. Support only the addition of the `client.connect` method
2. Show a deprecation warning if `client.connect` has not been called in advance
3. Raise an error if `client.connect` has not been called in advance

Each step will be released separately. And this can be handled separately from this PR, as it is a different concern.

koic:
Looks good to me! Can you squash commits into one?

lib/mcp/client/http.rb
```diff
@@ -105,7 +197,10 @@ def send_request(request:)
       # session state is cleared either way.
       # https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management
       def close
-        return unless @session_id
+        unless @session_id
```
atesgoral, line 200: @koic This was added since you reviewed:

```
  -        return unless @session_id
  +        unless @session_id
  +          clear_session
  +          return
  +        end
```

test/mcp/client/http_test.rb
```diff
@@ -633,6 +633,25 @@ def test_close_without_session_is_noop
         assert_nil(client.session_id)
       end
 
+      def test_close_clears_stateless_connection_state
```
atesgoral, line 636: @koic Corresponding test for that latest patch

lib/mcp/client/http.rb
koic: Thank you! I confirmed it using the "Compare" link on GitHub.

koic (approved):

Issue #334, opened 2026-05-02
Remove stdio's implicit init and require client.connect universally


atesgoral:
From discussion: https://github.com/modelcontextprotocol/ruby-sdk/pull/327#issuecomment-4323949406

> To minimize user impact while avoiding a breaking change, proceeding in the following steps looks appropriate:

- [x] Support only the addition of the client.connect method #336
- [x] Show a deprecation warning if client.connect has not been called in advance #338
- [x] Raise an error if client.connect has not been called in advance

koic:
It's my understanding that https://github.com/modelcontextprotocol/ruby-sdk/pull/327 does not implement `connect` for the stdio client. I've opened #336 as Phase 1.

koic:
I opened #427 to release this earlier than planned, so it can be included as one of the breaking changes before the v1 release.

PR #338, opened 2026-05-07
Warn on implicit stdio initialization

atesgoral:
## Summary

Implements phase 2 of #334 after the 0.15.0 release.

- Emit a deprecation warning when `MCP::Client::Stdio#send_request` performs the compatibility implicit initialization path.
- Keep the existing implicit initialization behavior for this release step.
- Cover the warning path and the explicit `MCP::Client#connect` no-warning path.
- Add an Unreleased changelog entry.

## How Has This Been Tested?

- `bin/rake test` -- 847 runs, 2159 assertions, 0 failures

## Notes

`bin/rake rubocop` currently reports existing unrelated `Minitest/AssertEmptyLiteral` offenses in files outside this change. No new offense is reported for the touched files.

### DIFF
--- a/lib/mcp/client/stdio.rb
+++ b/lib/mcp/client/stdio.rb
@@ -134,7 +134,10 @@ module MCP
 
       def send_request(request:)
         start unless @started
-        connect unless @initialized
+        unless @initialized
+          warn("Calling `MCP::Client::Stdio#send_request` without calling `MCP::Client#connect` is deprecated. Use `MCP::Client#connect` before sending requests instead.", uplevel: 1)
+          connect
+        end
 
         write_message(request)
         read_response(request)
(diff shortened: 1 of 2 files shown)


koic:
Thanks for the follow-up. Since v0.15.0 was just released, wait a bit before merging and releasing.

atesgoral:
@koic Shall we get this in now?

koic:
The timing for merging and releasing this Phase 2 is a bit of a balancing act, but the idea was to merge it after roughly a month had passed since the release of https://github.com/modelcontextprotocol/ruby-sdk/releases/tag/v0.15.0, which already included Phase 1.

At the same time, delaying it too long would reduce the chance for existing users to become aware of the transition, so it probably makes sense to merge it fairly soon.

On the other hand, the final behavior change in Phase 3 will be a breaking change, so it may be better to leave a reasonably long gap after Phase 2 before shipping that one (perhaps around three months?).

In any case, it probably makes sense to include this in the next release. Thank you!

koic (approved):

PR #353, opened 2026-05-16
Add OAuth 2.1 client support for MCP authorization flow

koic:
## Motivation and Context

The MCP Authorization spec (2025-11-25) requires clients to support Protected Resource Metadata discovery (RFC 9728), Authorization Server Metadata discovery (RFC 8414), Dynamic Client Registration, and OAuth 2.1 Authorization Code grant with PKCE (S256). The Ruby SDK previously had no client-side OAuth support: the HTTP transport raised `RequestHandlerError` on a 401 with no recovery, and 21 `auth/*` scenarios were listed as known failures in `conformance/expected_failures.yml`.

This aligns the Ruby SDK with the Python and TypeScript SDKs, both of which expose a pluggable OAuth client provider that drives the full discovery + DCR + authorization + token exchange flow. `MCP::Client::OAuth::Provider` mirrors that surface: callers supply `client_metadata`, a `redirect_uri`, and pluggable `redirect_handler` / `callback_handler` hooks plus a token/client information storage. `MCP::Client::OAuth::InMemoryStorage` is provided as a default.

`MCP::Client::HTTP` accepts a new optional `oauth:` keyword. When present, the transport attaches `Authorization: Bearer` to every request, and on a 401 parses `WWW-Authenticate` for the `resource_metadata` URL and `scope` challenge, drives the discovery + DCR + authorization code flow through the provider, and retries the failed request once with the acquired token.

Spec compliance details:

- The `WWW-Authenticate` parser handles multiple challenges (e.g. `Basic ..., Bearer ...` or `Bearer ..., DPoP ...`), extracting only the parameters that belong to the Bearer challenge.
- The PRM `resource` value is required to cover the MCP server URL (same scheme/host/port, with the PRM path as a prefix of the server path) before the flow continues, preventing redirection of issued tokens to a different audience.
- Both authorization and token requests always include a `resource` parameter (RFC 8707), defaulting to a canonicalized form of the MCP server URL when PRM does not advertise one.
- `client_secret_basic`, `client_secret_post`, and `none` token endpoint authentication methods are supported. When pre-registered client information specifies no `token_endpoint_auth_method`, the transport defaults to `client_secret_basic` per RFC 6749 §2.3.1 an dthe TS/Python SDK convention. Pre-registered client credentials supplied via `Storage#save_client_information` with either string or symbol keys are honored, skipping DCR.
- Scope resolution follows the spec order: the `WWW-Authenticate` challenge scope wins first, then PRM `scopes_supported`, and `Provider#scope` is the last-resort fallback only if neither is supplied (the provider-supplied scope must not pre-empt server-advertised scope).
- Token endpoint error responses (RFC 6749 §5.2) are parsed: a 4x with `error: "invalid_grant"` raises `Flow::InvalidGrantError`, so `MCP::Client::HTTP` can discard the stored refresh token only when the AS truly says it is dead. Generic `AuthorizationError` (5xx, network failures, other error codes) leaves the refresh token intact, so a transient AS outage does not force interactive reauth on the next attempt.
- The OAuth `state` parameter is compared via `OpenSSL.fixed_length_secure_compare` (with a `bytesize` guard) to prevent timing-based discovery of the expected value.
- The 401 retry guard is a request-scoped local variable, so an MCP server that keeps returning 401 is surfaced as an error after one retry rather than triggering an infinite re-authorization loop.

## How Has This Been Tested?

Unit tests cover PKCE generation, multi-challenge `WWW-Authenticate` parsing, PRM and Authorization Server metadata URL builders, the full Flow orchestrator (success path, state mismatch, missing PKCE support, skipping DCR for both string and symbol-keyed preregistered client information, `client_secret_basic` token endpoint authentication, PRM `resource` mismatch rejection, fallback to the server URL when PRM omits `resource`, scope resolution preferring PRM `scopes_supported` over `Provider#scope`, default `client_secret_basic` for confidential clients without an explicit `token_endpoint_auth_method`, and the new `InvalidGrantError` vs generic `AuthorizationError` split for token endpoint rejections), and the HTTP transport behavior (401 recovery + retry, no-provider fallthrough, no infinite loop on repeated 401, no infinite loop on flow failure, refresh-token retention on transient 5xx from the token endpoint).

Conformance: `bundle exec rake conformance` newly passes 16 `auth/*` scenarios beyond the existing baseline, including
`auth/metadata-default`, `metadata-var{1,2,3}`,
`scope-from-www-authenticate`, `scope-from-scopes-supported`, `scope-omitted-when-undefined`, `scope-retry-limit`, `token-endpoint-auth-{basic,post,none}`, `pre-registration`, `resource-mismatch`, and `offline-access-not-supported`. Remaining `auth/*` scenarios stay in `expected_failures.yml`. `bundle exec rake test rubocop` is clean.

## Breaking Changes

None. `MCP::Client::HTTP#initialize` accepts a new optional `oauth:` keyword that defaults to `nil`, preserving existing behavior. When unset, the transport raises `MCP::Client::RequestHandlerError` on a 401 exactly as before.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1885,6 +1885,113 @@ client.tools # will make the call using Bearer auth
 
 You can add any custom headers needed for your authentication scheme, or for any other purpose. The client will include these headers on every request.
 
+#### OAuth 2.1 Authorization
+
+When an MCP server enforces the [MCP Authorization spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization),
+pass an `MCP::Client::OAuth::Provider` to the transport instead of a static `Authorization` header. The transport will:
+
+- Send `Authorization: Bearer <access_token>` on every request when a token is available.
+- On a `401 Unauthorized`, parse the `WWW-Authenticate` header, discover the authorization server (Protected Resource Metadata + RFC 8414 Authorization Server Metadata),
+  perform Dynamic Client Registration if needed, run the OAuth 2.1 Authorization Code flow with PKCE (S256), and retry the failed request with the acquired token.
+- On subsequent 401s with a saved `refresh_token`, exchange it at the token endpoint before falling back to the full interactive flow (RFC 6749 Section 6).
+
+```ruby
+require "mcp"
+
+provider = MCP::Client::OAuth::Provider.new(
+  client_metadata: {
+    client_name: "My MCP App",
+    redirect_uris: ["http://localhost:3030/callback"],
+    grant_types: ["authorization_code", "refresh_token"],
+    response_types: ["code"],
+    token_endpoint_auth_method: "none",
+  },
+  redirect_uri: "http://localhost:3030/callback",
+  redirect_handler: ->(authorization_url) {
+    # Send the user to the authorization URL - typically `Launchy.open(authorization_url)`
+    # or a manual `puts authorization_url` in CLI tools.
+  },
+  callback_handler: -> {
+    # Capture the redirect (for example, by running a small HTTP listener on
+    # `redirect_uri`) and return [code, state] from the query string.
+  },
+)
+
+transport = MCP::Client::HTTP.new(
+  url: "https://api.example.com/mcp",
+  oauth: provider,
+)
+client = MCP::Client.new(transport: transport)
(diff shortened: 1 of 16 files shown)


atesgoral (approved):
Big gap that's being closed!

PR #362, opened 2026-05-26
Allow skipping validation for schema

dvisockas:
## Motivation and Context
If you're building dynamic tool schemas - every time you build one it has to be validated as the `Schema` class calls `validate_schema!` in the initializer. This can get slow (taking up to ~100ms) for deeper schemas. 

This PR allows the user of the `Schema` class to build the schema themselves and not pay the performance penalty of `JSON.parse(JSON.dump(schema))` and `validate_schema!`.

## How Has This Been Tested?
Added unit tests.

## Breaking Changes
None.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
An alternative

### DIFF
--- a/lib/mcp/tool/schema.rb
+++ b/lib/mcp/tool/schema.rb
@@ -13,10 +13,14 @@ module MCP
 
       attr_reader :schema
 
-      def initialize(schema = {})
-        @schema = JSON.parse(JSON.dump(schema), symbolize_names: true)
-        @schema[:type] ||= "object"
-        validate_schema!
+      def initialize(schema = {}, validate = true)
+        if validate
+          @schema = JSON.parse(JSON.dump(schema), symbolize_names: true)
+          @schema[:type] ||= "object"
+          validate_schema!
+        else
+          @schema = schema
+        end
       end
 
       def ==(other)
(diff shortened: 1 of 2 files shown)


koic:
Thanks for the PR and the feedback. I agree that the issue here is the performance hit from validating the schema on every instantiation.

Adding an option like `validate: false` (`validate = false`) is one way to speed this up, but it also allows invalid schemas to be created. And once a `validate` option is exposed, it becomes part of the public interface and is not easy to remove later.

So rather than skipping `validate_schema!` entirely, it seems better to improve performance by caching validation results for identical schemas.

I've opened #363 with the caching approach. What do you think?

dvisockas:
> Thanks for the PR and the feedback. I agree that the issue here is the performance hit from validating the schema on every instantiation.
> 
> Adding an option like `validate: false` (`validate = false`) is one way to speed this up, but it also allows invalid schemas to be created. And once a `validate` option is exposed, it becomes part of the public interface and is not easy to remove later.
> 
> So rather than skipping `validate_schema!` entirely, it seems better to improve performance by caching validation results for identical schemas.
> 
> I've opened #363 with the caching approach. What do you think?

Makes sense. Will leave comments on your opened PR if I see something

PR #369, opened 2026-06-01
Speed Up `Tool::Schema` Validation by 5x to 100x

### LINKED ISSUE #364: Inter-process schema validation cache
https://github.com/modelcontextprotocol/ruby-sdk/pull/363 introduced a in-process schema validation cache. The problem with this approach is that when the MCP server boots - the cache is stale and needs to be warmed, which might not always be trivial to do. 

Could we extract the default cache implementation and allow users to configure a different cache store, say for an inter-process long lived cache?

```ruby
MCP.configuration do |config|
  config.schema_validation_cache = FileBasedSchemaCache.new
end
```

Happy to put up a PR just want to align on the approach first

koic:
## Motivation and Context

`MCP::Tool::Schema` previously used the pure-Ruby `json-schema` gem for both construction-time metaschema validation and runtime argument / result validation. For deep schemas this cost ~32-100ms per construction (see issue #364).

`json_schemer` (https://rubygems.org/gems/json_schemer, v2.5.0) is an actively maintained alternative that is much faster on this workload. Switching reduces cold-start warming pressure in horizontally scaled multi-tenant deployments (the scenario raised in #364) and brings runtime data validation down to sub-millisecond per call once the per-instance schemer is memoized.

The implementation preserves the legacy behavior of the `json-schema` based code. Metaschema is pinned to draft-04 (`meta_schema:` option); the `$schema` dialect URI is still emitted in `to_h` as 2020-12 but runtime validation continues to use draft-04, exactly as before. Moving the runtime validator itself to 2020-12 is out of scope for this PR and warrants its own discussion. `format` keywords are not enforced (`format: false` option), matching what `json-schema`'s draft-04 path did. Malformed schemas (e.g. an invalid `pattern` regex) continue to surface as `ArgumentError, "Invalid JSON Schema: ..."`; the `RegexpError` `json_schemer` raises during construction is wrapped. `Symbol` values in argument data are coerced to strings inside the internal `stringify` helper so they validate against `type: "string"` the same way they did before.

The public `Schema#schema` accessor is dropped. It was a refactor artifact from #198 (the consolidation of `InputSchema`'s `properties` / `required` readers into a single hash) with no callers in `lib/`; tests that needed the merged hash now read it through `to_h`. The accessor's only remaining role would have been to expose a mutation path that could desynchronize the memoized `@schemer`, which this change removes.

Runtime dependency delta: drops `json-schema` (and `addressable`), adds `json_schemer`, `hana`, `regexp_parser`, and `simpleidn`. Each added gem is a single-purpose library tied to JSON Schema spec compliance.

Closes #364.

## How Has This Been Tested?

Side-by-side benchmark on representative schemas (simple, with `$ref` / `additionalProperties`, nested, depth 20, depth 40): construction-time metaschema validation is 4.7x to ~100x faster; runtime data validation with the memoized schemer is sub-millisecond across all sizes.

Four existing tests had their stub targets or message assertions updated to track the new validator while preserving the original test intent. The "unexpected errors bubble up" tests in `input_schema_test.rb` and `output_schema_test.rb` now stub `JSONSchemer::Schema#validate` instead of `JSON::Validator.fully_validate`. The cache tests in `schema_test.rb` now stub `JSONSchemer::Schema#validate_schema`. The two "detailed error message" tests in `tool_test.rb` switched from asserting `json-schema`'s exact wording to format-agnostic substrings (`"properties/count/minimum"`, `"number"`). The "required arguments are converted to strings" test in `input_schema_test.rb` now reads the result through `to_h[:required]` instead of the removed `schema` accessor.

Regression tests added for the legacy-behavior preservations above: `format` keyword is not enforced, invalid `pattern` raises `ArgumentError` (not `RegexpError`), and `Symbol` values validate against `type: "string"`.

## Breaking Changes

`Schema#schema` is no longer a public method. The accessor had no callers outside the test suite, but anyone reading the internal representation directly (rather than through `to_h`) will need to switch. `Schema` / `InputSchema` / `OutputSchema` otherwise keep their constructor signatures, `to_h` output, `ValidationError` / `ArgumentError` raise points, and the `$schema` dialect emission. The validator error message wording inside `ArgumentError` and `ValidationError` now comes from `json_schemer` and still includes the JSON pointer path and a description of the mismatch.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/tool/input_schema.rb
+++ b/lib/mcp/tool/input_schema.rb
@@ -12,9 +12,9 @@ module MCP
       end
 
       def missing_required_arguments(arguments)
-        return [] unless schema[:required].is_a?(Array)
+        return [] unless @schema[:required].is_a?(Array)
 
-        (schema[:required] - arguments.keys.map(&:to_s))
+        (@schema[:required] - arguments.keys.map(&:to_s))
       end
 
       def validate_arguments(arguments)
--- a/lib/mcp/tool/schema.rb
+++ b/lib/mcp/tool/schema.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require "digest"
-require "json-schema"
+require "json_schemer"
 
 module MCP
   class Tool
@@ -38,11 +38,10 @@ module MCP
 
       # JSON Schema 2020-12 is the default dialect for MCP schema definitions
       # per MCP 2025-11-25 (SEP-1613). Note: emission only — runtime validation
-      # is still performed against the JSON Schema draft-04 metaschema because
-      # the `json-schema` gem does not yet support 2020-12.
+      # is still performed against the JSON Schema draft-04 metaschema.
       JSON_SCHEMA_2020_12_URI = "https://json-schema.org/draft/2020-12/schema"
 
-      attr_reader :schema
+      DRAFT4_META_SCHEMA_URI = "http://json-schema.org/draft-04/schema#"
 
       def initialize(schema = {})
         @schema = JSON.parse(JSON.dump(schema), symbolize_names: true)
@@ -51,7 +50,7 @@ module MCP
       end
 
       def ==(other)
-        other.is_a?(self.class) && schema == other.schema
+        other.is_a?(self.class) && @schema == other.instance_variable_get(:@schema)
       end
 
       def to_h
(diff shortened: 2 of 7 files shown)


atesgoral (approved):
🚀

NobodysNightmare:
Hinting at a breaking change:

If previous code looked like `@tool_context.output_schema.validate_result(result.to_json)`, then this is now not working anymore, as apparently a "hashified" version is necessary. In our case (which is probably the worst case), this means: `JSON.parse(result.to_json)`, since we have no `to_h` available for the results.

PR #396, opened 2026-06-10
Extract `MCP::Client` pagination loop into `fetch_all_pages`

baxter2:
`#tools`, `#resources`, `#resource_templates`, and `#prompts` each repeated the same loop to follow `next_cursor` and collect every page, differing only in the `list_*` method called and the field read from each page. Move that loop into a single private `fetch_all_pages` helper so the four methods become one-liners and the cursor-cycle guard lives in one place.

Behavior is unchanged, the existing `MCP::Client` list tests cover it.

## Motivation and Context
`MCP::Client#tools`, `#resources`, `#resource_templates`, and `#prompts` each followed the server's pagination the same way: start with no cursor, request a page, collect its items, follow `next_cursor`, and stop when the cursor is absent or repeats. The four methods were identical apart from which `list_*` method they called and which field they read from each page, so the same loop, including the guard against a server that returns a repeated or cyclic cursor was written out four times. (The four whole-collection methods were introduced alongside pagination in #320.)

This moves that loop into a single private `fetch_all_pages` helper. Each of the four methods becomes a one-line call describing what it fetches, and the cursor-following logic now lives in one place instead of four.

## How Has This Been Tested?
`rake test` and `rake rubocop` pass locally. The behavior is unchanged the cursor loop is moved into the helper as written. So the existing `MCP::Client` list tests continue to cover these methods.

## Breaking Changes
None. `#tools`, `#resources`, `#resource_templates`, and `#prompts` keep the same signatures and return the same values; only their shared internals were extracted into a private helper.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/lib/mcp/client.rb
+++ b/lib/mcp/client.rb
@@ -146,21 +146,7 @@ module MCP
     #   end
     def tools
       # TODO: consider renaming to `list_all_tools`.
-      all_tools = []
-      seen = Set.new
-      cursor = nil
-
-      loop do
-        page = list_tools(cursor: cursor)
-        all_tools.concat(page.tools)
-        next_cursor = page.next_cursor
-        break if next_cursor.nil? || seen.include?(next_cursor)
-
-        seen << next_cursor
-        cursor = next_cursor
-      end
-
-      all_tools
+      fetch_all_pages { |cursor| list_tools(cursor: cursor) }.flat_map(&:tools)
     end
 
     # Returns a single page of resources from the server.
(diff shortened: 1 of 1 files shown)


lib/mcp/client.rb
```diff
@@ -380,6 +324,27 @@ def ping
...
+    # items concatenated. The `seen` set guards against a server that repeats or
+    # cycles cursors, so the loop always terminates.
+    def fetch_all_pages(items_method)
+      items = []
+      seen = Set.new
+      cursor = nil
+
+      loop do
+        page = yield(cursor)
+        items.concat(page.public_send(items_method))
```
koic: Since there is no expectation of unknown methods being invoked here, `public_send` seems a bit over-abstracted. A direct method call would be simpler and avoid the runtime overhead of `public_send`.

baxter2: Thank you for the feedback. 

Removed `public_send` in favor of direct calls, updated in the latest push.

koic:
This looks good to me. Can you squash your commits into one?

baxter2:
Squashed into a single commit 👍

koic (approved):

PR #406, opened 2026-06-14
Deprecate roots, sampling, and logging per SEP-2577

### LINKED ISSUE #390: SEP-2577: Deprecate Roots, Sampling, and Logging
Tracking implementation of [SEP-2577](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2577) for the 2026-07-28 MCP spec release.

atesgoral:
# TL;DR
Deprecate MCP Roots, Sampling, and Logging when protocol version `2026-07-28` or a later protocol version is negotiated or explicitly configured.

# Context
SEP-2577 deprecates Roots, Sampling, and Logging as of the `2026-07-28` protocol version while keeping them fully functional during the deprecation window.

Closes https://github.com/modelcontextprotocol/ruby-sdk/issues/390

# Changes
- Add a shared protocol deprecation helper; `2026-07-28` and later protocol versions trigger the deprecation warnings.
- Warn on deprecated server APIs and session-scoped calls for roots, sampling, and logging.
- Warn client transports when deprecated roots or sampling capabilities are used at or after the deprecation threshold.
- No warnings when the negotiated or configured protocol version is older, or when none has been established.
- Add YARD `@deprecated` notes and regression tests for older, threshold, future, and never-established protocol versions.

# Not covered (possible follow-up)
The SEP-2575 stateless/modern lifecycle (https://github.com/modelcontextprotocol/ruby-sdk/pull/475, https://github.com/modelcontextprotocol/ruby-sdk/pull/476) landed after this PR was shaped:
- Stateless HTTP: legacy RPCs sent with a `2026-07-28` `MCP-Protocol-Version` header run against ephemeral sessions that do not carry the header version, so they do not warn.
- Modern envelope: `logging/setLevel` and client `notifications/roots/list_changed` no longer exist in that lifecycle, so usage-site warnings largely do not apply; the deprecated per-request `logLevel` meta key emits no warning.

Warnings fail silent (never spurious) on these paths. Happy to wire the envelope/header version through in a follow-up if wanted.

# Testing
- `ruby -c` for all changed Ruby files
- Direct protocol-threshold and warning smoke checks with `ruby -Ilib`
- `git diff --check`
- Full test and RuboCop coverage via CI; the local gem environment cannot currently satisfy the bundle


atesgoral:
@koic Happy to push this through or were you thinking of doing the deprecation later?

koic:
I'd lean toward doing the deprecation later rather than merging the full behavior change now.

This SEP introduces a behavioral switch tied to protocol versioning, and I think that design makes sense because it preserves compatibility with existing protocol versions. However, the 2026-07-28 specification hasn't actually been released yet, and it doesn't appear to have been added to `SUPPORTED_STABLE_PROTOCOL_VERSIONS` in the Python or TypeScript SDKs either.

Given that, I'd prefer to wait until the 2026-07-28 release is closer before enabling the deprecation behavior.

In the meantime, it might be reasonable to land the advisory annotations on their own.

atesgoral:
@koic Sounds good! I can split out the advisory.

atesgoral:
Split the advisory annotations out into #429 so they can land independently of the protocol-version behavior change. I'll keep this PR open for the `2026-07-28`-gated runtime warnings and rebase it once that release is closer.

atesgoral:
@koic Are we ready for this?

koic:
I plan to release MCP Ruby SDK v1.0 tomorrow (#470). I will merge this PR once both Ruby SDK v1.0 and the MCP 2026-07-28 specification have been released. Thank you for your patience.

koic:
I've merged https://github.com/modelcontextprotocol/ruby-sdk/pull/476. I think it would make sense to include this change in v1.2.0 instead.

Could you please rebase this PR onto the latest `main` branch?

Also, since changelog entries are added at release time, it might be better not to include one here to avoid merge conflicts.

lib/mcp/protocol_deprecations.rb
```diff
@@ -0,0 +1,50 @@
...
+        "2026-07-28 (SEP-2577). Use stderr or OpenTelemetry instead."
+
+    MESSAGES = {
+      roots: ROOTS_MESSAGE,
+      sampling: SAMPLING_MESSAGE,
+      logging: LOGGING_MESSAGE,
+    }.freeze
+
+    def deprecated_roots_sampling_logging?(protocol_version)
+      protocol_version == Configuration::ROOTS_SAMPLING_LOGGING_DEPRECATED_PROTOCOL_VERSION
```
koic: It looks like this should support not only the 2026-07-28 specification, but also future specification versions.

atesgoral: Expanding it

koic:
Thanks!

PR #444, opened 2026-07-08
Add Rails example app

### LINKED ISSUE #44: Rails example
A simple rails example with some tools and resources would be immensely helpful

koic:
## Motivation and Context

The README.md documents the "Rails (mount)" integration pattern, but the repository had no runnable Rails example. This adds a minimal Rails application under examples/rails that mounts StreamableHTTPTransport at /mcp and demonstrates class-based tools and a resource with a read handler.

The MCP server is built in config/routes.rb rather than in an initializer because Zeitwerk sets up autoloading after initializers run, so the tool classes in app/tools cannot be referenced there.
Code reloading is disabled because the transport holds the server built at boot. The app is booted with puma instead of rackup because rackup inserts `Rack::Lint` in its development default, which rejects the capitalized response header names that the transport currently emits.

Closes #44.

## How Has This Been Tested?

The curl walkthrough in examples/rails/README.md has been verified end to end against the running app: initialize and the initialized notification, tools/list, tools/call for both tools, resources/list, resources/read, the standalone SSE stream, and session termination via DELETE.

## Breaking Changes

None. This only adds an example application and documentation links.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ end
 the [MCP Streamable HTTP transport spec](https://modelcontextprotocol.io/specification/latest/basic/transports#streamable-http),
 so no additional route configuration is needed.
 
+A complete runnable application using this approach is available in [`examples/rails`](examples/rails).
+
 ##### Rails (controller)
 
 While the mount approach creates a single server at boot time, the controller approach creates a new server per request.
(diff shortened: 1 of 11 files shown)


atesgoral (approved):
This will be super helpful!

PR #457, opened 2026-07-14
Fix RuboCop offenses detected with rubocop-shopify 3.0.0

koic:
## Motivation and Context

This repository does not track Gemfile.lock, so `bundle install` resolves to the latest RuboCop gems. Upgrading rubocop-shopify from 2.x to 3.0.0 surfaces 8 new offenses because the shared config no longer disables `Style/MutableConstant`, newly disables `Lint/RescueException` and `Naming/AccessorMethodName`, and drops `AllowedPatterns` for `Naming/MethodName`. Upgrading rubocop itself to 1.88 surfaces one more offense from the new `Style/ReduceToHash` cop.

This change fixes them as follows:

- Freeze mutable constants. Regexp constants are also flagged because `TargetRubyVersion` is 2.7, where Regexp literals are not frozen.
- Remove inline disable directives that became redundant.
- Restore `AllowedPatterns: ['\Atest_']` for `Naming/MethodName` so `def test_` style test names can reference class names such as `NoMethodError`.
- Swap out the frozen `SUPPORTED_STABLE_PROTOCOL_VERSIONS` constant in the protocol version fallback test since Mocha cannot stub methods on frozen objects.
- Use `to_h` instead of `each_with_object` in `Server#index_resources_by_uri`.

Ref: https://rubygems.org/gems/rubocop-shopify/versions/3.0.0

## How Has This Been Tested?

`bundle exec rake` passes: 1402 runs, 0 failures, 0 errors, and RuboCop reports no offenses. The conformance suite also passes against its baseline.

## Breaking Changes

None. Freezing these constants only affects code that mutates them in place, which has never been supported.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,3 +16,7 @@ Lint/IncompatibleIoSelectWithFiberScheduler:
 
 Minitest/LiteralAsActualArgument:
   Enabled: true
+
+Naming/MethodName:
+  AllowedPatterns:
+    - '\Atest_'
--- a/lib/json_rpc_handler.rb
+++ b/lib/json_rpc_handler.rb
@@ -16,7 +16,7 @@ module JsonRpcHandler
     PARSE_ERROR = -32700
   end
 
-  DEFAULT_ALLOWED_ID_CHARACTERS = /\A[a-zA-Z0-9_-]+\z/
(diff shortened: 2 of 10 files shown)


koic:
I'll merge this now so upcoming CI runs can pass.

PR #194, opened 2025-12-07
Show exception messages for tool call errors

koic:
## Motivation and Context

Claude Code sometimes only shows `Error: MCP error -32603: Internal error` when an unexpected exception occurs during a tool call.

This change broadens the exception handling scope so the server can surface the exception message more often, making failures easier to diagnose.

## How Has This Been Tested?

Existing tests pass.

## Breaking Changes

None.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #486, opened 2026-08-06
fix: stop leaking exception messages to clients via JSON-RPC error data

### LINKED ISSUE #485: Unexpected tool exceptions leak internal details to clients via JSON-RPC error `data`
## Summary

When a tool handler raises an unexpected exception, the exception message is sent verbatim to the MCP client in the JSON-RPC error response's `data` member.

```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"Internal error","data":"Internal error calling tool list_entity_types: undefined local variable or method 'foo' for class Tools::ListEntityTypes"}}
```

## Where (v1.0.0)

- `MCP::Server#call_tool` wraps any unexpected exception into `RequestHandlerError.new("Internal error calling tool #{tool_name}: #{e.message}", ...)`
- `JsonRpcHandler.handle_request_error` falls back to `error.message` for the `data` member, so the embedded exception message reaches the client
- The generic `rescue StandardError => e` in `JsonRpcHandler.process_request` likewise responds with `data: e.message`

## Why it matters

Exception messages routinely contain internals: class, method and variable names, and — for HTTP client errors raised inside tools (e.g. Faraday) — internal hostnames and URLs. For publicly reachable servers these leak to untrusted clients. The SDK already avoids this on its other error paths: the generic request-handler wrap uses a fixed `"Internal error handling #{method} request"` without the exception message, and the StreamableHTTP transport's catch-all answers with a plain `"Internal server error"` — `call_tool` is the one place that embeds `e.message`. This is a classic information-exposure-through-error-messages weakness ([CWE-209](https://cwe.mitre.org/data/definitions/209.html)).

## Suggested fix

Drop the `: #{e.message}` suffix from `call_tool`'s wrap, so the client-facing message (and the `data` member echoing it) is just `"Internal error calling tool #{tool_name}"` — the tool name is useful to the caller and safe to expose. The generic `rescue StandardError` fallback in `JsonRpcHandler.process_request` should likewise stop echoing `e.message` into `data`; capability errors (`"Server does not support ..."`), which currently rely on that fallback for their client-facing message, keep it by re-raising through `RequestHandlerError` — the channel whose message is deliberately surfaced. Observability is unaffected: the original exception already reaches `configuration.exception_reporter` via `original_error` before the response is built.

jj-ivx:
Unexpected exceptions raised by tool handlers — and anything reaching `JsonRpcHandler`'s blind `rescue StandardError` — had their `#message` echoed verbatim into the JSON-RPC error `data` member, exposing internals such as class, method and host names to untrusted clients ([CWE-209](https://cwe.mitre.org/data/definitions/209.html)).

- `Server#call_tool` no longer appends `e.message` to the wrapped `RequestHandlerError`; clients see `"Internal error calling tool <name>"`. Observability is unaffected: the original exception still reaches `configuration.exception_reporter` via `original_error`.
- `JsonRpcHandler.process_request`'s generic `rescue StandardError` no longer sets `data` (JSON-RPC 2.0 makes it optional).
- Capability errors keep their client-facing message (`"Server does not support ..."`) by re-raising through `RequestHandlerError` — the one channel whose message is deliberately surfaced to clients.

Fixes #485

### DIFF
--- a/lib/json_rpc_handler.rb
+++ b/lib/json_rpc_handler.rb
@@ -135,11 +135,12 @@ module JsonRpcHandler
       success_response(id: id, result: result)
     rescue MCP::Server::RequestHandlerError => e
       handle_request_error(e, id, id_validation_pattern)
-    rescue StandardError => e
+    rescue StandardError
+      # The exception message is deliberately not echoed into `data`: it can carry
+      # internals that must not reach untrusted clients (CWE-209).
       error_response(id: id, id_validation_pattern: id_validation_pattern, error: {
         code: ErrorCode::INTERNAL_ERROR,
         message: "Internal error",
-        data: e.message,
       })
     end
   end
--- a/lib/mcp/server.rb
+++ b/lib/mcp/server.rb
@@ -560,7 +560,14 @@ module MCP
         return
       end
 
-      Methods.ensure_capability!(method, capabilities)
+      begin
+        Methods.ensure_capability!(method, capabilities)
+      rescue Methods::MissingRequiredCapabilityError => e
+        # Re-raise through RequestHandlerError, the one channel whose message is
+        # deliberately surfaced to clients: `JsonRpcHandler`'s blind `rescue StandardError`
+        # no longer echoes exception messages into the error `data` member (CWE-209).
+        raise RequestHandlerError.new(e.message, request, error_type: :internal_error, original_error: e)
+      end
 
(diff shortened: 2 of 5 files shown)


koic:
For context, this behavior was originally introduced in #194 to improve the developer experience. That said, I think this change is appropriate now. Exposing the message could be made opt-in through configuration for development purposes, but that can be addressed separately if the need arises.

PR #494, opened 2026-08-08
Derive `Mcp-Name` from the body in the modern transport test helper

koic:
## Motivation and Context

#490 and #492 merged independently green but broke each other on main: #492 made the modern path reject a name-bearing POST whose body carries a target name without `Mcp-Name` (-32020, HTTP 400) and taught the `modern_rack_request` helper to derive `Mcp-Method` from the body, while #490, merged in between, added five modern `tools/call` tests that build their requests through that helper and therefore send no `Mcp-Name`. Since the #492 merge every CI run on main fails those five tests with 400 where 200 is expected; the library behavior itself is correct.

The helper now derives `Mcp-Name` the same way it derives `Mcp-Method`: from `params.name` or `params.uri` when the body's method is one of the name-bearing three. That is what a conforming client sends per the 2026-07-28 specification, and what the TypeScript SDK's client, the bundled `MCP::Client::HTTP`, and the conformance harness's standard headers all do, so tests built through the helper model a compliant client by default. `name_header: nil` omits the header for tests of the requirement itself, symmetric with `method_header: nil`, and an explicit `headers:` entry still overrides the derived value, which keeps the base64 mismatch tests exercising their divergent names.

## How Has This Been Tested?

The five failing tests pass again without being touched, and the #492 requirement tests (missing `Mcp-Method`, missing `Mcp-Name`, header mismatches, base64 decoding) still pass: the transport test file reports 217 runs with zero failures. `bundle exec rake` is green, including RuboCop and the `--requirements 2025-11-25` conformance baseline.

## Breaking Changes

None. The change is confined to a test helper; the shipped gem is untouched.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/test/mcp/server/transports/streamable_http_transport_test.rb
+++ b/test/mcp/server/transports/streamable_http_transport_test.rb
@@ -5357,6 +5357,7 @@ module MCP
         test "modern POST without the Mcp-Name header on a name-bearing method returns 400 with -32020" do
           response = @transport.handle_request(modern_rack_request(
             modern_body("tools/call", { name: "some_tool", arguments: {} }),
+            name_header: nil,
           ))
 
           assert_equal 400, response[0]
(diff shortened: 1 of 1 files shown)


koic:
Merging via bypass to restore the CI as quickly as possible.

PR #548, opened 2026-09-09
fix(client/oauth): omit body client_id when using HTTP Basic

junyuanz1:
Token requests using HTTP Basic currently also send `client_id` in the form body. Some authorization servers interpret that combination as multiple client authentication methods and reject the request with `invalid_request`.

Omit the body `client_id` when authenticating with Basic, including when Basic is selected by the existing fallback for stored credentials without an explicit authentication method. Keep body credentials for `client_secret_post`, client identification for unauthenticated requests, and JWT assertions for `private_key_jwt` in separate branches.

[RFC 6749 §3.2.1](https://www.rfc-editor.org/rfc/rfc6749.html#section-3.2.1) permits a token request to include `client_id`; it does not prohibit sending it alongside Basic. This is a compatibility fix that avoids the redundant identifier. Comments distinguish that rationale from the one-authentication-method requirement in §2.3.

Validation on Ruby 3.4.5:

- Full test suite: 1,781 tests, 4,658 assertions, no failures or errors.
- Regression coverage for client credentials, authorization-code exchange, and refresh: all three fail against upstream main and pass with the fix.
- Conformance: server 73 passed; client 452 passed with 9 failures covered by the existing baseline. Both baseline checks passed.
- RuboCop: 30 existing offenses, verified identical to upstream main; no new findings.
- `git diff --check` passed.

### DIFF
--- a/lib/mcp/client/oauth/flow.rb
+++ b/lib/mcp/client/oauth/flow.rb
@@ -997,9 +997,9 @@ module MCP
           post_to_token_endpoint(as_metadata: as_metadata, client_info: client_info, form: form)
         end
 
-        # Submits a form-encoded request to the token endpoint, applying
-        # the client authentication method advertised in `client_information` and
-        # adding `client_id` (and `client_secret` when not using HTTP Basic).
+        # Submits a form-encoded token request using the authentication method
+        # stored in `client_information`. The method determines whether client
+        # credentials belong in the form body, a Basic header, or a JWT assertion.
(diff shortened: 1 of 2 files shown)


junyuanz1:
this fixed the problem for the Notion MCP which doesn't allow `HTTP Basic credentials in the header plus client_id in the body`

koic:
@junyuanz1 Can you squash your commits into one?

junyuanz1:
> @junyuanz1 Can you squash your commits into one?

squashed

PR #555, opened 2026-09-17
Let a provider add parameters to the token requests it makes

### LINKED ISSUE #554: `client_credentials` token requests can't carry authorization-server-specific parameters
## Problem

`ClientCredentialsProvider` builds a fixed token request. In `Flow#run_client_credentials!` (v1.5.1):

```ruby
form = { "grant_type" => "client_credentials" }
effective_scope = resolve_scope(scope: scope, prm: prm)
authorize_request!(...)
form["scope"] = effective_scope if effective_scope
form["resource"] = resource if resource
```

and `post_to_token_endpoint` merges only `client_id` (or the `private_key_jwt` assertion). A provider has no way to add anything.

Our authorization server requires two extra parameters identifying the calling application and its environment. Because we can't add them, we can't use `oauth:` at all.

## Why that costs more than the token request

Opting out of `oauth:` opts out of everything else keyed on it:

- `session_headers` sets `Authorization` from `@oauth.access_token` on every request
- `send_request` rescues `Faraday::UnauthorizedError` and retries, gated on `@oauth`

The second matters more than it looks. `send_request` builds the per-request `SSEStream` as the first statement inside its `begin`, so `retry` starts a fresh buffer. A consumer who can't use `oauth:` must put the 401 retry elsewhere — and the obvious place, a Faraday middleware, is wrong. The middleware retries through the same `env`, whose `on_data` is still the same `SSEStream#on_data`. The 401 body stays in the buffer, the retried response is appended to it, and `parse_json_buffer` gets two concatenated JSON documents.

We shipped exactly that and it broke every call once tokens began needing refresh. It reads like a truncated response but is the opposite: `json_ensure_eof` firing on trailing bytes after a complete document.

So the missing hook doesn't just block a parameter — it steers consumers toward a retry placement that quietly corrupts the stream buffer.

## Proposal

Let a provider contribute extra token-request parameters:

```ruby
form.merge!(@provider.additional_token_params) if @provider.respond_to?(:additional_token_params)
```

`respond_to?`-guarded exactly like `client_assertion` already is, so existing providers are unaffected. Reserved keys could be rejected or simply win.

## Alternative

A broader version: let a provider own token acquisition outright. Some consumers already have a token service with its own caching and rotation and just want the SDK to call it. Today the only way in is overriding the private `run_oauth_flow!`.

## Current workaround

Skip `oauth:`, subclass the transport, and retry around the public `send_request` so each attempt rebuilds the stream:

```ruby
def send_request(request:)
  retried = false
  begin
    super
  rescue MCP::Client::RequestHandlerError => e
    raise unless e.error_type == :unauthorized && !retried
    retried = true
    raise unless refresh_token_from(e)
    retry
  end
end
```

Public API only, but it reimplements the SDK's own retry, and since the token never reaches `session_headers` we set the header on the connection ourselves via the customizer block from #303.

Happy to open a PR for whichever shape you prefer.

koic:
## Motivation and Context

`ClientCredentialsProvider` builds a fixed `client_credentials` token request: `grant_type`, `scope`, `resource`, and the client authentication that `Flow#post_to_token_endpoint` adds. An authorization server that requires further parameters, such as Auth0's `audience`, cannot be used through `oauth:` at all, and giving up `oauth:` also gives up the bearer header and the 401-driven retry that the transport provides.

RFC 6749 Section 8.2 allows extension parameters on the token request, and the TypeScript SDK lets a provider shape its token request through `prepareTokenRequest`. `ClientCredentialsProvider.new` now takes `token_request_params:`, a Hash of String keys and values added to every token request the provider makes, and `Flow#post_to_token_endpoint` reads the same name by duck typing, so a `Provider` or `CrossAppAccessProvider` subclass that defines the method gets the same treatment on its authorization code exchange, refresh, or `jwt-bearer` request.
`Provider.new` and `CrossAppAccessProvider.new` take the same keyword, held by `StorageBackedProvider` like `authorization_request_validator`, so every bundled provider refuses a bad value before anything is stored.

The parameters go underneath the flow's own, so the SDK's values always win, and a key the SDK sets itself (`Flow::RESERVED_TOKEN_REQUEST_PARAMS`) is refused rather than silently overridden: with `ArgumentError` from the constructor, before any `client_information` is written, and again with `ArgumentError` from the flow, before the token request is sent. Only Strings are accepted because `URI.encode_www_form` encodes other scalars and Arrays in ways the caller did not write, and the provider keeps a frozen copy so a later change to the caller's Hash cannot alter what is sent.
A Hash comparing keys by identity is refused too, since two equal keys are two entries there, sent twice by a provider method or silently collapsed into one by the copy. The flow raises `ArgumentError` rather than `AuthorizationError` because `MCP::Client::HTTP` treats a failed refresh as a reason to run the interactive flow, which would then fail the same way after the user signed in. Both refusals raise `Flow::InvalidTokenRequestParamsError`, a subclass of `ArgumentError`, so a caller can tell them from an unknown keyword, and the copy duplicates keys as well as values, since `Hash` copies only keys whose class is exactly `String`.

Fixes #554.

## How Has This Been Tested?

New tests cover every token request the parameters can ride (the `client_credentials` grant under all three client authentication methods, the authorization code exchange, refresh, and `jwt-bearer`), both refusal boundaries (the constructors of all three providers, and the flow for a provider defining the method), and the transport surfacing a refused hook from a refresh attempt without starting the interactive flow. All of them fail against the previous library.

## Breaking Changes

None. The keyword is optional and the hook is opt-in; a provider that does not define `token_request_params` sends the same requests as before.

## Types of changes

- [x] Documentation update

## Checklist


kstevens715:
Thanks for turning this around so fast — `token_request_params:` covers what we needed, and duck-typing it in `post_to_token_endpoint` is a better shape than what I proposed.

One adjacent thing, for a separate change rather than this one. `MCP::Client::HTTP` builds the flow with `Flow.new(provider: @oauth)` at http.rb:850 and :870, so there's no way to reach the `http_client_factory:` that `Flow` already accepts. We record outbound HTTP into per-turn traces for incident debugging, and our own token service instruments its connection today, so moving to `oauth:` would drop discovery and token requests out of those traces. Once this lands, that's the last thing keeping us off `oauth:`.

Filed as #558 — happy to open a PR for it.

koic:
Thanks for the offer. I've opened PR #560. This lands the capability in a different shape: `http_client_customizer:` on the provider rather than a factory on the transport, so the SDK keeps its connection defaults and can refuse a request that a customizer's middleware would send to another origin.

A recorder goes in as `http_client_customizer: ->(faraday) { faraday.use MyApp::HttpRecorder }`. The "Customizing the OAuth HTTP Client"  section of `docs/_client/authorization.md` in that PR has the details and the constraints a customizer takes on.

PR #573, opened 2026-09-23
Support a two-leg authorization code flow for web-hosted clients

### LINKED ISSUE #572: Support a two-leg authorization code flow for web-hosted clients
## Problem

`OAuth::Flow#run!` runs the whole authorization code flow in one call. It generates the PKCE verifier and `state` as locals, hands the authorization URL to `redirect_handler`, and then blocks on `callback_handler` for the code (`lib/mcp/client/oauth/flow.rb` at 1.6.0):

https://github.com/modelcontextprotocol/ruby-sdk/blob/6da3009a6ecd47c58312fa81c96c1ff2d693fc18/lib/mcp/client/oauth/flow.rb#L248-L283

`Provider.new` requires `callback_handler:`, and `MCP::Client::HTTP` drives `run!` synchronously on a `401` or a `403 insufficient_scope`. That works for CLI and desktop clients, where one process can open a browser and wait on a loopback listener.

It does not work for a client hosted in a web application. There, authorization starts in one HTTP request and the redirect lands on another, often in a different process, minutes later. Nothing is left to block on, and the verifier and `state` exist only on the stack of the first request.

Assembling the flow outside the SDK is not a good workaround. `Discovery` and `PKCE` are public, but client registration, authorization server metadata validation, issuer binding (SEP-2352), RFC 9207 `iss` validation, the token exchange, and issuer-stamped token storage are all private to `Flow`. A web client would have to reimplement the security-critical half of the flow to get around one blocking call.

## Use case

Our case is a hosted agent where each end user connects their own remote MCP servers. The user starts authorization from one request and the authorization server redirects to our callback endpoint, which is served by whichever pod receives it. Tokens and client registrations are persisted per user connection, which the existing `storage:` duck type already supports. What's missing is a way to persist the in-flight authorization and finish it later.

## Prior art

Two Tier 1 SDKs already split the flow:

- **TypeScript.** `auth()` returns `'AUTHORIZED' | 'REDIRECT'`. On the first leg it persists the verifier and discovery state through `saveCodeVerifier()` and `saveDiscoveryState()`, calls `redirectToAuthorization(url)`, and returns `'REDIRECT'`; the transport then throws `UnauthorizedError`. The callback leg is `transport.finishAuth(callbackParams)`, or `auth(provider, { serverUrl, authorizationCode, iss })`. It restores that state, validates `iss` against the recorded issuer, redeems the code at the recorded authorization server, and saves tokens. The docs require discovery state to be persisted "with the same durability as `codeVerifier`" so the callback leg is bound to the authorization server the redirect targeted.
  - [`auth.ts` (`return 'REDIRECT'`)](https://github.com/modelcontextprotocol/typescript-sdk/blob/60321700871029401a2e3bed8fdf4f02c9ec3331/packages/client/src/client/auth.ts#L1428-L1430)
  - [`streamableHttp.ts` (`finishAuth`)](https://github.com/modelcontextprotocol/typescript-sdk/blob/60321700871029401a2e3bed8fdf4f02c9ec3331/packages/client/src/client/streamableHttp.ts#L893-L902)
  - [`docs/clients/oauth.md`](https://github.com/modelcontextprotocol/typescript-sdk/blob/60321700871029401a2e3bed8fdf4f02c9ec3331/docs/clients/oauth.md)
- **Rust.** `AuthorizationManager#get_authorization_url` stores a `StoredAuthorizationState` in a pluggable `StateStore`, keyed by the CSRF token. The state holds the verifier, expected issuer, whether `iss` is required, the requested scopes, and `created_at`. `exchange_code_for_token_with_issuer(code, csrf_token, iss)` loads the entry by `state` and validates the issuer, and only then deletes the entry. A forged callback carrying the right `state` therefore cannot burn the verifier the real callback needs. `StateStore` is documented for Redis or database backends (modelcontextprotocol/rust-sdk#614).
  - [`auth.rs` (`StateStore`)](https://github.com/modelcontextprotocol/rust-sdk/blob/a730321f4b4489c58ba0198ed314acb834b8e074/crates/rmcp/src/transport/auth.rs#L455)
  - [`auth.rs` (`exchange_code_for_token_with_issuer`)](https://github.com/modelcontextprotocol/rust-sdk/blob/a730321f4b4489c58ba0198ed314acb834b8e074/crates/rmcp/src/transport/auth.rs#L2098)

Python (`redirect_handler` + `callback_handler`), C# (`AuthorizationCallbackHandler`), and Go (`AuthorizationCodeFetcher`) are single-leg, like Ruby today.

## Proposal

Follow TypeScript's shape, since `Provider` was modeled on its `OAuthClientProvider`, and add Rust's binding of pending state to `state`. Names below are placeholders.

```ruby
provider = MCP::Client::OAuth::Provider.new(
  client_metadata: client_metadata,
  redirect_uri: "https://agent.example.com/oauth/mcp/callback",
  redirect_handler: ->(url) { hand_to_user(url) },
  # No callback_handler: the flow stops after the redirect.
  storage: storage, # also implements the pending-authorization methods below
)
flow = MCP::Client::OAuth::Flow.new(provider: provider)

# Request A: begin
flow.run!(server_url: server_url, resource_metadata_url: resource_metadata_url, scope: scope)
# => :redirect

# Request B, possibly another process, with a provider built over the same storage: finish
flow.finish!(server_url: server_url, callback_params: request.query_parameters)
# => :authorized
```

- **`callback_handler:` becomes optional.** Without it, `run!` persists a pending authorization, calls `redirect_handler`, and returns `:redirect`. Providers that pass `callback_handler:` keep the current behavior. Internally, one leg can become begin, then `callback_handler`, then finish, so both shapes share every check.
- **Pending state lives in `storage`, keyed by `state`.** Add optional `save_pending_authorization(state, pending)`, `pending_authorization(state)`, and `delete_pending_authorization(state)` methods, with `InMemoryStorage` implementing them. The two-leg path requires them. The pending entry records:
  - the PKCE verifier
  - a snapshot of the authorization server metadata (issuer, token endpoint, token auth method, `authorization_response_iss_parameter_supported`)
  - the client registration in use
  - `resource`, `redirect_uri`, and the requested scopes
  - `created_at`
- **`Flow#finish!` takes the whole callback query**, as TypeScript's `finishAuth(URLSearchParams)` does. Presence of `iss` then comes from the query itself, instead of from the length of the Array that `callback_handler` returns today.
- **`MCP::Client::HTTP` with `oauth:` and no `callback_handler`** raises a dedicated error after leg 1 instead of retrying. That error sits outside `AuthorizationError`, so the refresh-failure fallback does not catch it, and it exposes the authorization URL. Requests after `finish!` pick up the stored tokens. The `403 insufficient_scope` step-up path behaves the same way.

## Security requirements for the callback leg

1. Look the pending entry up by `state` before any network I/O, and reject an unknown or expired `state` without echoing callback parameters.
2. Validate `iss` against the recorded issuer, and require it when the recorded metadata advertises support, before consuming the entry. This follows Rust's ordering.
3. Surface the callback's `error` / `error_description` only after the issuer check passes, since they are attacker-controlled in a mix-up. This follows TypeScript.
4. Delete the pending entry after the issuer check and before the token request, so it is single-use.
5. Redeem the code at the recorded token endpoint with the recorded client registration, `resource`, and `redirect_uri`. Do not re-run discovery on the callback leg, so the code is redeemed at the authorization server the user was sent to (SEP-2352).
6. Enforce a maximum age from `created_at`, and make it configurable.
7. Never log the verifier; the pending object's `inspect` redacts it.

`authorization_request_validator`, `token_request_params:`, and `http_client_customizer:` keep applying on leg 1 and to the leg-2 token request exactly as they do today.

Binding the callback to the end user who started it stays with the application. For example, scoping `storage` per user means a callback delivered to another user's session cannot find the pending entry. The docs should say so explicitly.

## Open questions

- Should `finish!` also exist as a convenience on `MCP::Client::HTTP` (TypeScript has both), or does the `Flow` entry point suffice, since a callback endpoint usually has no live transport?
- What should the default maximum age of a pending entry be?

atesgoral:
Closes #572.

## Motivation and Context

`OAuth::Flow#run!` runs the whole authorization code flow in one call and blocks on `callback_handler` for the code, which suits CLI and desktop clients but not a client hosted in a web application: there, authorization starts in one HTTP request and the redirect lands on another, often in a different process. The verifier and `state` exist only on the first request's stack, and assembling the flow outside the SDK means reimplementing registration, issuer binding, `iss` validation, and the token exchange that are private to `Flow`.

This follows the TypeScript SDK's shape (`auth()` returning `'REDIRECT'`, then `finishAuth()`), which `Provider` was already modeled on, and adds the Rust SDK's binding of pending state to `state` (`StateStore`), consuming it only after the issuer check.

**With `callback_handler` (unchanged):** the flow blocks until the code comes back.

```mermaid
sequenceDiagram
  participant App
  participant HTTP as MCP::Client::HTTP
  participant Flow as OAuth::Flow
  participant User as User (browser)
  participant AS as Authorization server
  participant MCP as MCP server

  App->>HTTP: tools/list
  HTTP->>MCP: POST
  MCP-->>HTTP: 401 + WWW-Authenticate
  HTTP->>Flow: run!
  activate Flow
  Flow->>MCP: Protected Resource Metadata
  Flow->>AS: AS metadata, registration
  Note over Flow: PKCE verifier and state live on this stack
  Flow->>User: redirect_handler(authorization_url)
  rect rgba(255, 170, 0, 0.12)
    Note over Flow,AS: Flow blocks on callback_handler for as long as the user takes
    User->>AS: sign in and consent
    AS-->>User: redirect to redirect_uri
    User-->>Flow: callback_handler returns [code, state, iss]
  end
  Flow->>Flow: check state and iss
  Flow->>AS: token request (code, verifier)
  AS-->>Flow: tokens
  Flow-->>HTTP: :authorized
  deactivate Flow
  HTTP->>MCP: retry with Bearer token
  MCP-->>HTTP: 200
  HTTP-->>App: result
```

**Without `callback_handler`:** each leg returns, and the pending authorization lives in `storage` between them.

```mermaid
sequenceDiagram
  participant A as Request A
  participant HTTP as MCP::Client::HTTP
  participant FlowA as Flow (leg 1)
  participant Store as Shared storage
  participant User as User (browser)
  participant AS as Authorization server
  participant B as Request B (redirect_uri)
  participant FlowB as Flow (leg 2)

  rect rgba(80, 120, 255, 0.08)
    Note over A,FlowA: Leg 1, in whichever process serves request A
    A->>HTTP: tools/list
    Note over HTTP: MCP server answers 401
    HTTP->>FlowA: run!
    FlowA->>AS: discovery, registration
    FlowA->>Store: save_pending_authorization(state, pending)
    FlowA-->>HTTP: redirect_handler(url), then :redirect
    HTTP-->>A: raise AuthorizationPendingError(authorization_url)
    A-->>User: 302 to authorization_url
    Note over A: Request A ends, nothing waits
  end

  User->>AS: sign in and consent
  AS-->>User: redirect to redirect_uri

  rect rgba(80, 120, 255, 0.08)
    Note over B,FlowB: Leg 2, possibly another process
    User->>B: GET redirect_uri?code&state&iss
    B->>FlowB: finish!(server_url, callback_params)
    FlowB->>Store: pending_authorization(state)
    Note over FlowB: check age, server_url, endpoints, iss
    FlowB->>Store: delete_pending_authorization(state)
    FlowB->>AS: token request at recorded endpoint (code, verifier)
    AS-->>FlowB: tokens
    FlowB->>Store: save_tokens
    B-->>User: done
  end

  Note over A,HTTP: Later requests send the stored Bearer token
```

- `callback_handler:` becomes optional. Without it, `Flow#run!` saves a pending authorization in `storage` keyed by `state`, calls `redirect_handler`, and returns `:redirect`; the flow's `authorization_url` reader returns the URL. Providers that pass `callback_handler:` behave exactly as before.
- `storage` gains three optional methods, `save_pending_authorization(state, pending)`, `pending_authorization(state)`, and `delete_pending_authorization(state)`, required only without `callback_handler` (`Provider::PendingAuthorizationStorageError` otherwise). `InMemoryStorage` implements them. A pending authorization is a JSON-compatible Hash: the PKCE verifier, the canonical MCP server URL, `resource`, `redirect_uri`, `client_id`, the validated authorization server metadata, and `created_at`. The client secret is not copied into it.
- `Flow#finish!(server_url:, callback_params:)` takes the redirect's whole query, so the presence of `iss` comes from the query itself. It:
  1. looks the pending authorization up by `state` before any request (unknown, used, malformed, or oversized `state` is refused), and discards and refuses one older than `pending_authorization_max_age` (new `Provider` keyword, default 600 seconds);
  2. requires `server_url` to name the server the authorization began with, and re-checks the recorded endpoints, since the metadata made a round trip through application storage;
  3. validates `iss` against the recorded issuer, and requires it when the recorded metadata advertises support, before consuming the entry, so a forged callback with a valid `state` cannot burn the real verifier;
  4. consumes the entry, then surfaces an `error` response (bounded like token endpoint errors) only after the issuer check;
  5. redeems the code at the recorded token endpoint with the recorded client registration (a registration replaced in the meantime is refused), `resource`, and `redirect_uri`, without running discovery again (SEP-2352).
- `MCP::Client::HTTP` with such a provider raises `Flow::AuthorizationPendingError` after the first leg instead of retrying, on both the `401` and the `403 insufficient_scope` step-up paths. It sits outside `AuthorizationError` so the refresh fallback does not catch it, and it exposes `authorization_url` while keeping the URL (which carries `state`) out of the message.

Binding the callback to the user who started the authorization stays with the application; the docs call this out and recommend per-user storage.

Open questions from #572, left as proposed here:

- `finish!` lives on `Flow` only. A callback endpoint usually has no live transport, so I did not add a convenience on `MCP::Client::HTTP`; happy to add one if we want parity with TypeScript's `transport.finishAuth`.
- The default maximum age of a pending authorization is 600 seconds.

## How Has This Been Tested?

- New `test/mcp/client/oauth/two_leg_flow_test.rb` drives both legs through a storage that round-trips every value through JSON, with a fresh provider and flow on the second leg standing in for another process. It covers the saved entry and the PKCE commitment, the token request made from the recorded state without rediscovery, single use, unknown/missing/oversized/Array `state`, expiry, a different MCP server, a replaced registration, a CIMD `client_id`, tampered recorded endpoints, a malformed entry, a missing code, and a one-leg provider. On `iss`, it checks that a mismatch keeps the entry (and the legitimate callback then completes), that a missing `iss` is refused when advertised, that an error response is surfaced after a matching `iss`, and that an error response with a mismatched `iss` is not surfaced.
- `test/mcp/client/oauth/http_oauth_test.rb`: a `401` with a two-leg provider raises `AuthorizationPendingError` without retrying, and the next request after `finish!` succeeds with the stored token.
- `test/mcp/client/oauth/provider_test.rb`: optional `callback_handler`, the storage requirement, `pending_authorization_max_age` validation, and `InMemoryStorage`.
- Mutation check: moving the entry deletion ahead of the `iss` check fails three of the new tests.
- `bundle exec rake test` (1913 runs, 0 failures) and `bundle exec rake rubocop` pass locally on Ruby 3.4.8.

## Breaking Changes

None. `callback_handler:` moves from required to optional; providers that pass it, and custom storages used with them, are unaffected.

## Types of changes

- [x] Documentation update

## Checklist


lib/mcp/client/oauth/flow.rb
```diff
@@ -449,6 +488,80 @@ def refresh!(server_url:, resource_metadata_url: nil)
...
+          # through the application's storage.
+          ensure_secure_endpoints!(as_metadata, server_url: pending["server_url"])
+
+          validate_authorization_response_issuer!(
+            as_metadata: as_metadata,
+            iss: callback_param(callback_params, "iss"),
+            iss_provided: true,
+          )
+
+          @provider.delete_pending_authorization(state)
```
koic, line 544: The read in `pending_authorization` and the delete here are separate calls, so two callbacks with the same `state` arriving close together (a retried redirect) can both pass every check and redeem the code twice. Reproduced with a storage that sleeps 200 ms in `pending_authorization`: two concurrent `finish!` calls both returned `:authorized`. The second exchange is then denied and the first one's tokens may be revoked.

Since the storage contract is new, it is worth closing now: have `delete_pending_authorization(state)` return the entry it removed (`nil` when there was none, as `Hash#delete` already does), and let `finish!` proceed only when it gets one back. The `iss`-before-consume ordering stays as it is.

lib/mcp/client/oauth/provider.rb
```diff
@@ -138,6 +180,25 @@ def initialize(
         def authorization_flow
           :authorization_code
         end
+
+        # True when the authorization finishes in a later request: the flow saves a pending authorization,
+        # hands the authorization URL to `redirect_handler`, and returns, and `Flow#finish!` completes it
+        # in the request that receives the redirect.
+        def two_leg?
```
koic, line 187: "Two-leg" collides with the OAuth community's vocabulary, where two-legged means the flow with no user involved and three-legged the user-authorization flow this PR splits. A reader who knows that vocabulary will read it as the opposite of what the option does.

The mode may not need a name at all. `two_leg?` is `@callback_handler.nil?` and nothing more, so `Flow` can read `callback_handler` directly and the predicate can go, which also keeps one less name out of the public API; every authorization-code provider already responds to it. The docs heading can name the audience instead, such as "Authorization in Web Applications", and the test file can use the PR's own noun, `pending_authorization_flow_test.rb`; the PR title and changelog line would follow. `finish!`, `:redirect`, `AuthorizationPendingError`, and `pending_authorization` read fine as they are.

What do you think about that?

lib/mcp/client/oauth/flow.rb
atesgoral: Good catch, thanks. Done: `delete_pending_authorization(state)` now returns the entry it removed (`nil` when there was none), and `finish!` proceeds only when it gets a well-formed entry back. The `iss`-before-consume ordering is unchanged.

`finish!` checks for an entry rather than any truthy value so that a storage returning a count, as Redis `DEL` does, fails closed (`0` is truthy in Ruby). The docs now ask for an atomic remove-and-return (`GETDEL` in Redis, `DELETE ... RETURNING` in SQL).

I added a test that holds two concurrent `finish!` calls at a barrier after their reads: one returns `:authorized`, the other is refused, and the token endpoint sees a single request.

lib/mcp/client/oauth/provider.rb
atesgoral: Agreed, and thanks for catching the vocabulary clash. I dropped `two_leg?`, and `Flow` now reads `callback_handler` directly. The docs section is now "Authorization in Web Applications", the test file is `pending_authorization_flow_test.rb`, and "two-leg" is gone from the code and docs. The PR title and changelog line now read "Let a web application finish authorization in the request that receives the redirect".

atesgoral: Also updated the language in #572.

